PR #25, opened 2024-03-08
Bug fix to retrieve the correct selected ice candidate pair

david-macpherson:
## Relevant components:
- [ ] Signalling server
- [x] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
The active candidate pair returned from the `AggregatedStats.getActiveCandidatePair` function is returning the 1st candidate pair element which has received any bytes, some browsers will receive data on multiple candidate pairs which is not the selected candidate pair

## Solution
Implement the handling of the [`RTCTransportStats`](https://developer.mozilla.org/en-US/docs/Web/API/RTCTransportStats) from the [`RTCStatsReport`](https://developer.mozilla.org/en-US/docs/Web/API/RTCStatsReport)

Retrieve the selected candidate pair from the list of candidate pairs using the [`selectedcandidatepairid`](https://developer.mozilla.org/en-US/docs/Web/API/RTCTransportStats#selectedcandidatepairid) from the [`RTCTransportStats`](https://developer.mozilla.org/en-US/docs/Web/API/RTCTransportStats)

Due to [`browser_compatibility`](https://developer.mozilla.org/en-US/docs/Web/API/RTCTransportStats#browser_compatibility) fire fox doesn't support [`RTCTransportStats`](https://developer.mozilla.org/en-US/docs/Web/API/RTCTransportStats) 

In the case of the [`RTCTransportStats`](https://developer.mozilla.org/en-US/docs/Web/API/RTCTransportStats) being undefined  fall back to use the [`selected`](https://developer.mozilla.org/en-US/docs/Web/API/RTCIceCandidatePairStats/selected) property of the [`RTCIceCandidatePairStats`](https://developer.mozilla.org/en-US/docs/Web/API/RTCIceCandidatePairStats)

### DIFF
--- a/Frontend/library/src/PeerConnectionController/AggregatedStats.ts
+++ b/Frontend/library/src/PeerConnectionController/AggregatedStats.ts
@@ -323,7 +323,7 @@ export class AggregatedStats {
         // Check if the RTCTransport stat is not undefined
         if (this.transportStats){
             // Return the candidate pair that matches the transport candidate pair id
-            return this.candidatePairs.find((candidatePair) => candidatePair.id = this.transportStats.selectedCandidatePairId, null);
+            return this.candidatePairs.find((candidatePair) => candidatePair.id === this.transportStats.selectedCandidatePairId, null);
         }
         
         // Fall back to the selected candidate pair
--- a/Frontend/library/src/PeerConnectionController/AggregatedStats.ts
+++ b/Frontend/library/src/PeerConnectionController/AggregatedStats.ts
@@ -33,6 +33,7 @@ export class AggregatedStats {
     sessionStats: SessionStats;
     streamStats: StreamStats;
     codecs: Map<string, string>;
+    transport: RTCTransportStats;
 
     constructor() {
         this.inboundVideoStats = new InboundVideoStats();
@@ -94,6 +95,7 @@ export class AggregatedStats {
                     this.handleTrack(stat);
                     break;
                 case 'transport':
+                    this.handleTransport(stat);
                     break;
                 case 'stream':
                     this.handleStream(stat);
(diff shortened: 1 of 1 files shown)


Frontend/library/src/PeerConnectionController/AggregatedStats.ts
```diff
@@ -33,6 +33,7 @@ export class AggregatedStats {
     sessionStats: SessionStats;
     streamStats: StreamStats;
     codecs: Map<string, string>;
+    transport: RTCTransportStats;
```
lukehb: Consider renaming the field `transportStats` to be consistent with the other fields here.

Frontend/library/src/PeerConnectionController/AggregatedStats.ts
```diff
@@ -312,6 +319,14 @@ export class AggregatedStats {
      * @returns The candidate pair that is currently receiving data
      */
     public getActiveCandidatePair(): CandidatePairStats | null {
-        return this.candidatePairs.find((candidatePair) => candidatePair.bytesReceived > 0, null)
+        
+        // Check if the RTCTransport stat is not undefined
+        if (this.transport){
+            // Return the candidate pair that matches the transport candidate pair id
+            return this.candidatePairs.find((candidatePair) => candidatePair.id = this.transport.selectedCandidatePairId, null);
```
lukehb: After squinting at my screen and reading this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find#callbackfn

I _think_ you what you want here is:

`return this.candidatePairs.find((candidatePair) => candidatePair.id === this.transport.selectedCandidatePairId, null);`

lukehb: Note doing a comparison instead of an assignment.

lukehb (changes requested):
Minor changes requested to logic and naming.

lukehb (approved):
lgtm

PR #46, opened 2024-03-26
Add flag to enable flex-fec support

Belchy06:
## Relevant components:
- [ ] Signalling server
- [X] Frontend library
- [X] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
WebRTC has supported receiving Flexible Forward Error Corrected (FlexFec) streams as default since M92. With PixelStreaming recently adding support for this feature, it makes sense for the frontend to support it as well.

## Solution
This PR adds a new flag (and corresponding UI toggle) for enabling FlexFec support. With the toggle enabled, the browser will signal its support for FlexFec to the Unreal Engine application. If the application also supports FlexFec, the video stream will be protected.

## Documentation
Updated `Settings Panel.md` to contain information about the new toggle.

Tested connecting to streams with the toggle both enabled and disabled. If the application supports FlexFec and the toggle is enabled, additional information regarding the FEC packets can be seen in Chrome's webrtc-internals.

### DIFF
--- a/Frontend/library/src/PeerConnectionController/PeerConnectionController.ts
+++ b/Frontend/library/src/PeerConnectionController/PeerConnectionController.ts
@@ -407,6 +407,26 @@ export class PeerConnectionController {
                             });
                         });
 
+                    // If the user has toggle the "UseFlexFec" flag, we want to add it to the top of the codec preferences list.
+                    // We don't want it as a selectable codec because the first "actual" codec in the list is the codec that will be
+                    // protected.
+                    if(this.config.isFlagEnabled(Flags.UseFlexFec)) {
+                        const matcher = /(flexfec).*/;
+                        RTCRtpReceiver.getCapabilities('video').codecs.forEach((codec) => {
+                            const str = codec.mimeType.split('/')[1] + ' ' + (codec.sdpFmtpLine || '');
+                            const match = matcher.exec(str);
+                            if (match !== null) {
+                                codecs.push({
+                                    mimeType: codec.mimeType,
+                                    clockRate: codec.clockRate,
+                                    sdpFmtpLine: codec.sdpFmtpLine
+                                        ? codec.sdpFmtpLine
+                                        : ''
+                                });
+                            }
+                        });
+                    }
+
                     for (const codec of codecs) {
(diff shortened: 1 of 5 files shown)


Frontend/library/src/PeerConnectionController/PeerConnectionController.ts
```diff
@@ -407,6 +407,26 @@ export class PeerConnectionController {
...
+                                codecs.push({
+                                    mimeType: codec.mimeType,
+                                    clockRate: codec.clockRate,
+                                    sdpFmtpLine: codec.sdpFmtpLine
+                                        ? codec.sdpFmtpLine
+                                        : ''
+                                });
+                            }
+                        });
+                    }
```
lukehb, line 428: Given this whole code path is only evaluated if `transceiver.setCodecPreferences` exists, what happens on FF, does flexfec work?

Frontend/library/src/PeerConnectionController/PeerConnectionController.ts
```diff
@@ -407,6 +407,26 @@ export class PeerConnectionController {
                             });
                         });
 
+                    // If the user has toggle the "UseFlexFec" flag, we want to add it to the top of the preferences list.
+                    // We don't want it as a selectable codec because the first "actual" codec in the list is the codec that will be
+                    // protected.
+                    if(this.config.isFlagEnabled(Flags.UseFlexFec)) {
+                        const matcher = /(flexfec).*/;
```
lukehb, line 414: To save ourselves having to write something similar when we want to add ulpfec of RED support, can we add all the capabilities to our codec preferences?

lukehb (changes requested):
Some questions about what happens on FireFox and also a question about supporting all the capabilities in a more general way.

If we can support all the capabilities in a more general way, my preference would be to do that and in fact remove the new setting from the settings panel and just let the browser send everything it has (with our preferred one on top).

lukehb:
Bad bot

lukehb:
Made redundant by: https://github.com/EpicGamesExt/PixelStreamingInfrastructure/pull/248

PR #86, opened 2024-04-23
Added SDP for probator

### LINKED ISSUE #85: [BUG] - The resolution is not back to normal in Simulcast.
**UE Version:**
UE 5.3

**Frontend Version:**
UE5.3-0.3.0

**Problem component**
SFU

**Description**
When using Simulcast in SFU, the resolution changes to a lower resolution but does not return to the original resolution even if bandwidth is increased.

**Steps to Reproduce:**
1. Launch app with Simulcast option. (-SimulcastParameters=1.0,800000,2000000,2.0,100000,800000,4.0,10000,100000)
2. Reduce the bandwidth and see if the resolution goes down.
3. The resolution does not return when back to the original bandwidth.

**Expected behavior**
 When the bandwidth is back, the resolution is also back to normal.

**Screenshots**
simulcast
(I use Android 14 due to bandwidth limitations)

 - OS: Windows 11 Pro
 - Browser + version Android14 chrome 113

zuvola:
## Relevant components:
- [ ] Signalling server
- [x] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [x] SFU

## Problem statement:
The current SFU does not always select the appropriate resolution stream even when simulcast is configured.
This may be because mediasoup is sending a stream for inspection under the name "probator", but it is not receiving this stream properly, so the inspection is not performed properly and the bitrate estimation is not done properly.

## Solution
Fixes #85 

simul

### DIFF
--- a/Frontend/library/src/VideoPlayer/StreamController.ts
+++ b/Frontend/library/src/VideoPlayer/StreamController.ts
@@ -31,7 +31,8 @@ export class StreamController {
             'handleOnTrack ' + JSON.stringify(rtcTrackEvent.streams),
             6
         );
-        
+        // Do not add the track if the ID is `probator` as this is special track created by mediasoup for bitrate probing.
+        // Refer to https://github.com/EpicGamesExt/PixelStreamingInfrastructure/pull/86 for more details.
         if (rtcTrackEvent.track.id == 'probator') {
             return;
         }
--- a/Frontend/library/src/VideoPlayer/StreamController.ts
+++ b/Frontend/library/src/VideoPlayer/StreamController.ts
@@ -31,6 +31,11 @@ export class StreamController {
             'handleOnTrack ' + JSON.stringify(rtcTrackEvent.streams),
             6
         );
+        
+        if (rtcTrackEvent.track.id == 'probator') {
+            return;
+        }
+
         const videoElement = this.videoElementProvider.getVideoElement();
 
         if (rtcTrackEvent.track) {
--- a/SFU/mediasoup-sdp-bridge/lib/index.js
+++ b/SFU/mediasoup-sdp-bridge/lib/index.js
@@ -138,10 +138,14 @@ class SdpEndpoint {
         });
         const sendMsid = uuid_1.v4().substr(0, 8);
         console.log("[SdpEndpoint.createOffer] Make 'sendonly' SDP Offer");
+        var videoRtpParameters;        
         for (let i = 0; i < this.consumers.length; i++) {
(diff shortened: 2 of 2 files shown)


lukehb:
@zuvola Thanks for this digging any reference in mediasoup docs/discussion about this "probator"?

Also, when you remove this, does it fix the problem of not selecting the appropriate resolution stream?

lukehb:
Some further reading: https://mediasoup.discourse.group/t/probator-pli/3898/5

zuvola:
@lukehb 
Unfortunately I have not found any useful documentation.
The simulcast was working fine in the mediasoup demo at the following URL, and I found the probator when I was trying to find the difference.

https://v3demo.mediasoup.org/
https://github.com/versatica/mediasoup-demo

I just added the probator to the SDP using the mediasoup-client source as a reference and it works correctly, but I am not sure if this modification eliminates the root cause.
I am also not sure what happens when the probator packets are removed. Perhaps the bandwidth estimation will stop working.

lukehb:
With your PR applied, does automatic layer switching work when you restrict the bandwidth?

I do agree the probator does seem like a likely mechanism the mediasoup authors are using to probe for bitrate. 

> I am also not sure what happens when the probator packets are removed. Perhaps the bandwidth estimation will stop working.

I agree that bandwidth estimation will likely fail if probator is munged out. 

I think further investigation of the media soup demo is a good idea, perhaps we are configuring wrong.

zuvola:
By applying this PR, I have confirmed that it is working correctly, at least in my and my team member's environment.
It's a little hard to see, but I attached a working gif above.

Frontend/library/src/VideoPlayer/StreamController.ts
```diff
@@ -31,6 +31,11 @@ export class StreamController {
             'handleOnTrack ' + JSON.stringify(rtcTrackEvent.streams),
             6
         );
+
```
lukehb:
```suggestion
        // Do not add the track if the ID is `probator` as this is special track created by mediasoup for bitrate probing.
        // Refer to https://github.com/EpicGamesExt/PixelStreamingInfrastructure/pull/86 for more details.
```

lukehb (changes requested):
Added a minor suggestion to give some context to future readers and maintainers of the code.

lukehb (approved):
Had some discussions internally, this looks good to us. Thanks for finding this one and providing a neat and tidy PR. The Pixel Streaming user base and its maintainers thank you!

lukehb:
@mcottontensor To land and backport to 5.2, 5.3, 5.4 and make new releases.

PR #87, opened 2024-04-26
fix:kicked player should not try to reconnect

ibreathebsb:
## Relevant components:
- [x] Signalling server
- [x] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
kicked player still tries to reconnect

## Solution
1. change Signalling server to send code 3000 when player get kicked
2. change Frontend library websocket on close,if code is 3000 then don't reconnect

### DIFF
--- a/Frontend/library/src/WebRtcPlayer/WebRtcPlayerController.ts
+++ b/Frontend/library/src/WebRtcPlayer/WebRtcPlayerController.ts
@@ -220,9 +220,11 @@ export class WebRtcPlayerController {
             // https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent/code
             // lists all the codes. 
             const CODE_GOING_AWAY = 1001;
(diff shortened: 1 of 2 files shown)


lukehb:
@mcottontensor I do think this PR is a good concept. I think it likely needs to some updating for the 5.5 branch and I also think instead of doing an if check for a whole bunch of disconnect codes it might be better to store them in a set and do a DontReconnectCodes.Contains() or something like that. I suppose potentially we should also expose a bit of API so the user can add to the set of codes if they are doing a custom disconnect that should not trigger a reconnect.

lukehb:
On further review, with the move to UE 5.5 deprecating `kick` message this PR is no longer relevant for PSInfra versions going forward. CC @mcottontensor

PR #88, opened 2024-05-01
Adding an option to hide the UI.

mcottontensor:
## Relevant components:
- [ ] Signalling server
- [x] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
Sometimes it would be nice to hide the UI on the frontend.

## Solution
Added a config option `HideUI=true` to hide the UI on the frontend.


lukehb (approved):
Approved. Lgtm. I don't think we need to backport this one as it is a new feature not a bugfix.

PR #90, opened 2024-05-03
Updating build github actions

mcottontensor:
Updating the workflows.
- Moved the node version to a variable so it's easier to see/change. Probably wants to be a repo variable.
- Added a frontend build action to test that the frontend components build after any change
- Build checks run on any branch, not just master.

### DIFF
--- a/.github/workflows/run-frontend-build.yml
+++ b/.github/workflows/run-frontend-build.yml
@@ -2,10 +2,8 @@ name: Run frontend build
 
 on: 
   push:
-    branches: ['master']
     paths: ['Frontend/**']
   pull_request:
-    branches: ['master']
     paths: ['Frontend/**']
 
 env:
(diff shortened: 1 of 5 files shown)


.github/workflows/run-frontend-build.yml
```diff
@@ -0,0 +1,29 @@
+name: Run frontend build
+
+on: 
+  push:
+    paths: ['Frontend/**']
+  pull_request:
+    paths: ['Frontend/**']
+
+env:
+  NODE_VERSION: 18.17.x
```
lukehb, line 10: May as well get this one too.

lukehb (approved):
lgtm!

Yeah in 5.5 with /common folder maybe we can add a .NodeVersion file or something like that can be read by cli scripts and actions.

PR #113, opened 2024-05-08
Replaces first and last double quote sanitization which are added to the default launch arguments for http_root affecting startup

wywarren:
## Relevant components:
- [x] Signalling server
- [ ] Common library
- [ ] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
Without this fix, the current start server procedure won't resolve to the proper path of `player.html`. The server will report back with `Unable to locate file player.html`

## Solution
In `index.ts` we sanitize the merged program options using a replace regex to strip the leading and trailing double quotes from `http_root` if they exist.

## Screenshots
Current `master` error:
image

### DIFF
--- a/SignallingWebServer/src/index.ts
+++ b/SignallingWebServer/src/index.ts
@@ -97,6 +97,9 @@ if (!cli_options.no_config) {
 // merge the configurations
 const options: IProgramOptions = { ...config_file, ...cli_options };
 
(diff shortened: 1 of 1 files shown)


mcottontensor:
This should no longer be needed when #114 is merged.

PR #119, opened 2024-05-15
Fix: path normalizing on Windows

Belchy06:
## Relevant components:
- [ ] Signalling server
- [ ] Common library
- [ ] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [X] Platform scripts
- [ ] SFU

## Problem statement:
Path normalizing is broken in the Windows platform scripts.

image

## Solution
Removed superfluous quotes

### DIFF
--- a/SignallingWebServer/platform_scripts/cmd/common.bat
+++ b/SignallingWebServer/platform_scripts/cmd/common.bat
@@ -141,7 +141,7 @@ IF "%FRONTEND_DIR%"=="" (
 )
 
 rem try to make it an absolute path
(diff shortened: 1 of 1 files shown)


mcottontensor:
So without the quotes it fails to normalize paths with spaces if I recall correctly. Can you try putting the frontend in a path with a space in it?

lukehb:
@Belchy06 I did some checking today, @mcottontensor  is right this PR has issues if the path has spaces in it. I have prepared another PR here that fixes this and handles the spaces: https://github.com/EpicGamesExt/PixelStreamingInfrastructure/pull/122

PR #122, opened 2024-05-20
Fixed platform scripts on Windows looking in wrong frontend dir, sensitive to spaces in paths, and not setting user PATH to contain NodeJS.

lukehb:
## Relevant components:
- [ ] Signalling server
- [ ] Common library
- [ ] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [x] Platform scripts
- [ ] SFU

## Problem statement:

A few problems were identified:

1. The frontend directory which the Windows platform scripts would try to create/look for was unable to create a normalized path from combining two paths together.
2. A number of the `call` sites in the Windows platforms scripts were sensitive to directories which had a space in them.
3. When NodeJS did not exist on the system or user PATH, the platform scripts would fail as they only set NodeJS on the process PATH.
4. Powershell was being used in some cases, PowerShell is not guaranteed to be callable on a fresh install of non-pro versions of Windows without user intervention.

## Solution

Each of the issues was resolved as follows:

1. The frontend path was combined together by using/not using quotes in certain places to escape paths with spaces in them.
2. A number of the `call` functions in the scripts needed to be escaped with quotes.
3. The user PATH is now retrieved, and it is updated to include the NodeJS path. This is undone after the scripts have built everything. HOWEVER, the caveat to this change is Windows batch can only set the user PATH using `setx` - setx has a 1024 character limit. Therefore if the user PATH exceeds 1024 characters the script will emit a warning saying it will not set the path and the user can either trim their PATH or manually install NodeJS into the PATH.
4. Powershell calls were removed and replaced with windows batch equivalents.

## Documentation

No docs changes

- Test the platforms scripts on Windows with NodeJS not present on the system.
- Also tested using a path with a space in it. 
- Tested `start.bat --frontend-dir=D:\Pixel Streaming Infrastructure\SignallingWebServer\www` (note the spaces).
- Test all `start_*.bat` files with space in path

### DIFF
--- a/SignallingWebServer/platform_scripts/cmd/common.bat
+++ b/SignallingWebServer/platform_scripts/cmd/common.bat
@@ -157,7 +157,7 @@ for /f "tokens=2*" %%A in ('reg query "HKCU\Environment" /v PATH ^| find "PATH"'
 rem Store user's PATH in OLD_PATH
 set OLD_PATH=%USER_PATH%
 rem Update this processes PATH to have node dir (this WILL NOT persist outside this process which is a problem for NPM run scripts we use)
-set PATH="%NODE_DIR%;%PATH%"
+set PATH=%NODE_DIR%;%PATH%
 set PATH_LENGTH=0
 for %%A in ("%NODE_DIR%;%OLD_PATH%") do set "PATH_LENGTH=%%~zA"
 
--- a/SignallingWebServer/platform_scripts/cmd/common.bat
+++ b/SignallingWebServer/platform_scripts/cmd/common.bat
@@ -132,60 +132,85 @@ rem NOTE: We want to use our NodeJS (not system NodeJS!) to build the web fronte
 rem Save our current directory (the NodeJS dir) in a variable
 set NODE_DIR=%SCRIPT_DIR%node
 
-rem Prepend NODE_DIR to PATH temporarily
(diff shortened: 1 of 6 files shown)


SignallingWebServer/platform_scripts/cmd/common.bat
```diff
@@ -216,7 +237,7 @@ call :SetupCoturn
 exit /b
 
 :SetPublicIP
-FOR /f %%A IN ('powershell -command "(Invoke-Webrequest "http://api.ipify.org").content"') DO set PUBLIC_IP=%%A
+FOR /f %%A IN ('curl http://api.ipify.org') DO set PUBLIC_IP=%%A
```
mcottontensor, line 240: I did not realise windows now includes curl by default.

lukehb: Yeah, is cool! https://curl.se/windows/microsoft.html

lukehb:
I am happy with your most recent commit @mcottontensor

mcottontensor (approved):

lukehb:
This was subsequently simplified in: https://github.com/EpicGamesExt/PixelStreamingInfrastructure/commit/34977d86d83c0c0dbc10945346dded462c3a2fb1

PR #148, opened 2024-06-20
Fixing SFU probator stream exclusion. This will fix using SFU with Firefox

mcottontensor:
## Relevant components:
- [ ] Signalling server
- [ ] Common library
- [x] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
When using an SFU the stream never shows on Firefox.

## Solution
We were deliberately excluding a track called 'probator' that is added by mediasoup but on firefox the track id was not set and instead the stream id was set to probator. This change just changes where we look for the id.


lukehb (approved):
lgtm!

PR #149, opened 2024-06-20
Fixing frontend stats when using SFU

mcottontensor:
Peer connection now tracks the audio and video tracks and ignores the probator track, and then requests stats on each of the tracks individually.

## Relevant components:
- [ ] Signalling server
- [ ] Common library
- [x] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
When using the SFU there is an extra video track that resets the video stats to zero in Firefox since each tracks stats come in through the same path.

## Solution
We record the tracks for audio and video and ignore the probator track. We then request stats only for the recorded tracks.


lukehb (approved):
I am little mindful of storing the tracks as members - but I think it is okay for now (until we support say multiple audio tracks).

PR #167, opened 2024-07-03
Frontend - Provide default values to avoid potential NAN value returned when using dynamic player aspect ratios

richardFocus:
## Relevant components:
- [ ] Signalling server
- [ ] Common library
- [x] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
Our app uses various UI layouts for different pages and the pixel streaming player resizes to fit the available space as needed.  Currently there is an issue in the where the value of the X mouse coord in CoordinateConverter.tsx can return NAN.  I have found this to happen when the stream first starts and when the player hasn't yet resized and therefore could be zero, however resizing the player to zero width could also cause similar issues.

## Solution
The PR provides temporary default values that are divisible and prevents NAN from being returned.

## Documentation

What steps have you taken to ensure this PR maintains compatibility with the existing functionality? 
I have printed the returning value to the screen from every possible combination of layouts I have available.

### DIFF
--- a/Frontend/library/src/Util/CoordinateConverter.ts
+++ b/Frontend/library/src/Util/CoordinateConverter.ts
@@ -92,11 +92,12 @@ export class CoordinateConverter {
         this.videoElement = this.videoElementProvider.getVideoElement();
 
         if (this.videoElementParent && this.videoElement) {
-            const playerAspectRatio =
-                this.videoElementParent.clientHeight /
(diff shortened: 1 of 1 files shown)


lukehb (approved):
Thanks for finding this!

PR #197, opened 2024-07-16
Disconnect flow fixes

### LINKED ISSUE #176: [BUG] - Disconnect overlay incorrectly showing error overlay
**UE Version:**
N/A

**Frontend Version:**
UE5.4 (likely effects multiple versions)

**Problem component**
Frontend

**Description**
If you take a look at

https://github.com/EpicGamesExt/PixelStreamingInfrastructure/blob/0524416fea42ab00d0bfbd04fcb547e6c895ac50/Frontend/library/src/WebRtcPlayer/WebRtcPlayerController.ts#L243

The outcome of the boolean for the second argument is what decides whether or not a click to reconnect overlay should be displayed or an error overlay.

As a result of this logic, an 'error overlay' will be triggered on anything other than a 1001 close code. 1000 is a normal closure and should not result in an error overlay.

It also appears that the boolean to reconnect is tightly coupled with `MaxReconnectAttempts` which IMO should not be the case, as I would expect that to relate only to automatic reconnects. (Setting this flag to 0 should effectively disable automatic reconnects but not impact a users ability to manually click to restart the stream).

For our use case, we wanted the user to *always* be able to reconnect the stream (even if there was an abnormal closure code). And we also found a grammar error in the AFK text which we were able to fix at the same time.

```js
// override the _onDisconnect function to intercept webrtc disconnect events
// and modify how the event is fired by always showing a click to reconnect overlay.
// we also add a full stop to the AFK message.
stream._onDisconnect = function (eventString: string) {

	// check if the eventString coming in is the inactivity string and add a full stop
	if (eventString == "You have been disconnected due to inactivity") {
		eventString += "."
	}

	this._eventEmitter.dispatchEvent(
		new WebRtcDisconnectedEvent({
			eventString: eventString,
			allowClickToReconnect: true
		})
	);
}
```

 - OS: Ubuntu Linux 22.04 LTS
 - Chrome 126.0.6478.126 (Official Build) (64-bit)

mcottontensor:
## Relevant components:
- [ ] Signalling server
- [ ] Common library
- [x] Frontend library
- [x] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
Certain situations allowed the click to restart stream message to be omitted or added when they shouldn't be.

## Solution
Better handling of the disconnect/reconnect states

Fixes #176

### DIFF
--- a/Frontend/library/src/WebRtcPlayer/WebRtcPlayerController.ts
+++ b/Frontend/library/src/WebRtcPlayer/WebRtcPlayerController.ts
@@ -135,7 +135,7 @@ export class WebRtcPlayerController {
             this.onAfkTriggered.bind(this)
         );
         this.afkController.onAFKTimedOutCallback = () => {
-            this.closeSignalingServer('You have been disconnected due to inactivity');
+            this.closeSignalingServer('You have been disconnected due to inactivity.');
         };
 
         this.freezeFrameController = new FreezeFrameController(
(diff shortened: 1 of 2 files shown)


lukehb (approved):
lgtm

PR #199, opened 2024-07-17
The URL parameters should now be case insensitive in the frontend.

### LINKED ISSUE #12: Frontend settings in URL are case sensitive
**UE Version:**
UE 5.3.2

**Frontend Version:**
UE 5.3

**Problem component**
Frontend settings

**Description**
Setting frontend settings such as match viewport resolution, hovering mouse mode and lightmode are case sensitive in the URL.
For example, adding `/?HoveringMouse=true` to your URL works, but `/?hoveringmouse=true` does not.

**Steps to Reproduce:**
1. Start a basic local pixel stream
2. Add a lowercase frontend setting to the URL, e.g. /?hoveringmouse=true
3. Observe setting failing to set.

**Expected behavior**
Frontend settings should function as intended, regardless of casing in URL.

 - OS: Windows 10 + Ubuntu
 - Chrome, Firefox and Edge

**Additional context**

---

Moved from https://github.com/EpicGames/PixelStreamingInfrastructure/issues/430

mcottontensor:
## Relevant components:
- [ ] Signalling server
- [ ] Common library
- [x] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
The URL parameters are case sensitive.

## Solution
We now handle the URL parameters in a case insensitive way.

Fixes #12

### DIFF
--- a/Frontend/library/src/Config/SettingBase.ts
+++ b/Frontend/library/src/Config/SettingBase.ts
@@ -6,6 +6,8 @@
 export class SettingBase {
     id: string;
     description: string;
+    useUrlParams: boolean;
+    _urlParams: Record<string, string>;
     _label: string;
     _value: unknown;
     onChange: (changedValue: unknown, setting: SettingBase) => void;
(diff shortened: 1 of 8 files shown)


lukehb (approved):
lgtm!

PR #205, opened 2024-07-24
Stream test github action

mcottontensor:
Adding a github action to run a streamer in software rendering to make sure that the stream can still be started.

### DIFF
--- a/.github/workflows/healthcheck-streaming.yml
+++ b/.github/workflows/healthcheck-streaming.yml
@@ -33,10 +33,10 @@ jobs:
       with:
         repository: 'EpicGamesExt/PixelStreamingInfrastructure'
         tag: 'minimal-streamer'
-        fileName: 'Minimal-PixelStreamer.7z'
+        fileName: 'Minimal-PixelStreamer-5.5.7z'
 
     - name: Extract streamer
-      run: 7z x -oStreamer Minimal-PixelStreamer.7z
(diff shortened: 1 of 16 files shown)


.github/workflows/healthcheck-streaming.yml
```diff
@@ -0,0 +1,114 @@
...
+    - uses: actions/setup-node@v4
+      with:
+        node-version: 'v18.17.0'
+
+    - name: Download streamer
+      uses: robinraju/release-downloader@v1
+      with:
+        repository: 'EpicGamesExt/PixelStreamingInfrastructure'
+        tag: 'minimal-streamer'
+        fileName: 'Minimal-PixelStreamer.7z'
```
lukehb: Can we version this file? E.g. `Minimal-PixelStreamer-Win-5.4.7z`

Extras/MinimalStreamTester/Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM mcr.microsoft.com/windows:ltsc2019
```
lukehb: Maybe filename should be `Dockerfile-Win`?

lukehb (changes requested):
Lgtm. Minor recommendations about file/release names to be more explicit.

lukehb:
As an aside, let's add the healthcheck badges to our readme.md now that we have a few of them.

lukehb (approved):
lgtm!

PR #210, opened 2024-07-24
Simplifying windows path modification for node

mcottontensor:


lukehb (approved):
This works great, I just tested it in cmd.exe and powershell with no NodeJS/NPM in my path,

PR #212, opened 2024-07-25
Removing browser send offer support.

### LINKED ISSUE #164: Drop support for browser send offer for 5.5
**Component your feature request relates to**
- [ ] Frontend

**Problem your feature solves? Please describe.**
We should drop support for browser sending offer, which has not been the default since 4.27. This will significantly clean up codepaths/supported cases in the internal libraries/wrappers.

mcottontensor:
## Relevant components:
- [x] Signalling server
- [x] Common library
- [x] Frontend library
- [x] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

Removing the browser send offer support as requested in #164

### DIFF
--- a/Common/docs/classes/Logger_Logger.Logger.md
+++ b/Common/docs/classes/Logger_Logger.Logger.md
@@ -45,7 +45,7 @@ logging to the browser console.
 
 #### Defined in
 
-[Logger/Logger.ts:8](https://github.com/mcottontensor/PixelStreamingInfrastructure/blob/e96d9c6/Common/src/Logger/Logger.ts#L8)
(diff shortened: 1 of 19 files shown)


lukehb (changes requested):
Also requires a docs update here: 

https://github.com/EpicGamesExt/PixelStreamingInfrastructure/blob/master/Common/docs/messages.md#playerconnected

https://github.com/EpicGamesExt/PixelStreamingInfrastructure/blob/master/Common/docs/messages.md#answer

lukehb:
I guess given this is a breaking protocol change, we should also bump signalling protocol version 

Edit: Remove example of updating protocol, that was entirely the wrong spot.

lukehb (approved):

Issue #203, opened 2024-07-23
Game UI Occluded by Navigation Bar on Mobile


lcdlv:
## Component your question relates to
- [ ] Signalling server
- [x ] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] SFU
- [ ] This repo
- [ ] Other

## Background
My game UI is Occluded by navigation bar on Mobile.

## Question
How can I adjust the player to fill the mobile screen and prevent UI overlay?

Here's an example:

This is on Windows:
![Screenshot_1](https://github.com/user-attachments/assets/d4b01c71-d822-4e5c-aea5-b365d99b0718)

This is on Mobile:
![Screenshot_2](https://github.com/user-attachments/assets/e4596dff-c30a-48e6-a455-350e3d382068)

lukehb:
What if you enable the match viewport resolution option in the frontend?

lcdlv:
Hi @lukehb ! It's enabled right now. I'm using Signalling Server for Unreal 5.3.
I'm not specialist in web development, so I need you point me where to touch to change that behaviour.

lukehb:
Try changing in `https://github.com/EpicGamesExt/PixelStreamingInfrastructure/blob/master/Frontend/implementations/typescript/src/player.html`

`height: 100vh;` to `height: 100dvh;` (note the `d`)

Which should dynamically account for top and bottom on device nav showing: https://developer.mozilla.org/en-US/docs/Web/CSS/length#dynamic

If it works let me know and I will push this live.

lcdlv:
HI @lukehb !! That's works!!!
Thanks!!

![Screenshot_1](https://github.com/user-attachments/assets/27104341-1721-4772-986e-765779adfbf0)

PR #213, opened 2024-07-25
Fix: Mobile UI not scaling correctly when nav bars are present.

lukehb:
## Relevant components:
- [x] Examples

## Problem statement:
Mobile UI was not scaling properly when match viewport resolution was used and mobile nav bars were present.

See here clipped UI:

image

## Solution
Parent container for Pixel Streaming should use dynamic viewport height, e.g. `100dvh`

`dvh` units are documented here: https://developer.mozilla.org/en-US/docs/Web/CSS/length#dynamic

They are compliant with all modern browsers.

When this fix is applied above UI looks like this:

image

## Documentation
N/A

- Tested with customer to confirm fix
- Tested with Chrome + PC
- Tested with Firefox + PC
- Tested with Android + Chrome
- Tested with Safari + iPhone
- Tested with Chrome + iPhone

### DIFF
--- a/Frontend/implementations/typescript/src/player.html
+++ b/Frontend/implementations/typescript/src/player.html
@@ -21,7 +21,7 @@
 </head>
 
 <!-- The Pixel Streaming player fills 100% of its parent element but body has a 0px height unless filled with content. As such, we explicitly force the body to be 100% of the viewport height -->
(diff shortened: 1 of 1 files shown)


lukehb:
Fixes: https://github.com/EpicGamesExt/PixelStreamingInfrastructure/issues/203

lukehb:
After some more testing `dvh` was still occluded on Safari IPhone, `svh` seems to be working well for all browsers, including Safari.

mcottontensor (approved):

PR #218, opened 2024-08-01
Release archive excludes update.

mcottontensor:
Updated the archive exclusions to remove anything non source.
Also made a small change to the platform scripts to make sure the common library was built properly. It was failing otherwise.


lukehb (approved):
lgtm!

Issue #171, opened 2024-07-09
[BUG] - Video element showing  edit text


adityayaduvanshi:
**UE Version:**
E.g. UE 5.1.1

**Frontend Version:**
E.g. UE5.2-0.3.0

**Problem component**
Video element showing edit text.

**Description**
I implemented the pixel stream in react js. its  working fine but in the  mobile  view it shows this message. 
![WhatsApp Image 2024-06-04 at 1 56 51 PM](https://github.com/EpicGamesExt/PixelStreamingInfrastructure/assets/38696088/416942f1-0d7e-42d7-91db-ecfab0dd24ec)

adityayaduvanshi:
Any help?

mcottontensor:
Should be fixed with the latest change in master from #238

PR #238, opened 2024-08-11
Set default inline style for hidden input

Belchy06:
## Relevant components:
- [ ] Signalling server
- [ ] Common library
- [X] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
The "hidden input" html elements exist in the base library. However, they are styled from the UI-library. This means that if you have a simple implementation that uses just the base library without the UI library, you will be forced into having these elements on screen.

## Solution
Sets the default css attributes in the base library so that these elements will be hidden without needing to apply the styles from the UI library.


mcottontensor:
Fixes #171

mcottontensor (approved):

lukehb:
This should be backported to our supported versions too @mcottontensor

mcottontensor:
Done.

PR #240, opened 2024-08-16
Change --no-save to --save in Wilbur

lukehb:
## Relevant components:
- [x] Signalling server
- [ ] Common library
- [ ] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
In previous versions of the PSInfra the `config.json` file was not overridden with arguments passed at launch. The current behaviour means by default `config.json` changes as a user passes launch args. This change in behaviour has confused some users already.

## Solution
This PR changes the `--no-save` argument to be a `--save` argument, which is `false` by default. This means the `config.json` file will only be written when `--save` is specified. The side effect is this means a `config.json` will not be generated when the user intially runs `start.bat`. This makes discovery of the configuration options a little more hidden (because no one reads the docs apparently). To mitigate this discovery issue I have also committed a default `config.json` file (and subsequently `.gitignore`'d it so that it will not be committed by PSInfra devs during our work).

## Documentation
I have updated the docs regarding the `--no-save` becoming `--save`.

I have given this a few local tests on Windows, both with and without `config.json` present. I have not tested this on Linux/Mac, but I do not foresee any issues there as this is a pure NodeJS code change, so it should be cross platform.

### DIFF
--- a/SignallingWebServer/config.json
+++ b/SignallingWebServer/config.json
@@ -1,0 +1,21 @@
+{
+	"log_folder": "logs",
+	"log_level_console": "info",
+	"log_level_file": "info",
+	"streamer_port": "8888",
+	"player_port": "80",
+	"sfu_port": "8889",
+	"serve": true,
+	"http_root": "D:\\PixelStreamingInfrastructure\\SignallingWebServer\\www",
+	"homepage": "player.html",
+	"https": false,
+	"https_port": 443,
+	"ssl_key_path": "certificates/client-key.pem",
+	"ssl_cert_path": "certificates/client-cert.pem",
+	"https_redirect": true,
+	"rest_api": false,
+	"peer_options": "",
+	"log_config": true,
+	"stdin": false,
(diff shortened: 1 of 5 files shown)


SignallingWebServer/config.json
```diff
@@ -0,0 +1,21 @@
...
+	"https_port": 443,
+	"ssl_key_path": "certificates/client-key.pem",
+	"ssl_cert_path": "certificates/client-cert.pem",
+	"https_redirect": true,
+	"rest_api": false,
+	"peer_options": "",
+	"log_config": true,
+	"stdin": false,
+	"console_messages": "verbose"
+}
```
lukehb, line 21: Note: The `.gitignore` still ignores this file, so while it will it exist in the repo, changes to it will not trigger changed files in git (which would be annoying during dev).

mcottontensor (approved):

PR #244, opened 2024-08-20
Add logging about streamer disconnecting

lukehb:
## Relevant components:
- [x] Signalling server
- [ ] Common library
- [ ] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
There is no indication of the streamer disconnecting without debug logging. In previous versions, this was part of the standard logging.

## Solution
Restore streamer disconnect logging as part of the standard logging that occurs when you run start.bat

## Documentation
N/A

Tested locally on Windows and using this PR.


Belchy06 (approved):
LGTM!

PR #251, opened 2024-08-26
Add: Bidirectional video support

Belchy06:
## Relevant components:
- [ ] Signalling server
- [ ] Common library
- [X] Frontend library
- [X] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
PixelStreaming is currently only able to receive video from the Unreal Engine. WebRTC supports bidirectional video, but this functionality hasn't been added to the frontend. This functionality allows easier testing of the PSPlayer feature among other things.

## Solution
This PR adds support for sending the browser webcam back to the Unreal Engine for processing and display.

## Documentation
Updated all relevant docs.

Tested the PixelStreaming still works in both UseCamera states.


Belchy06:
NOTE: Type names and docs links will need to be updated once confirmed.

Frontend/library/src/PeerConnectionController/PeerConnectionController.ts
```diff
@@ -394,19 +398,59 @@ export class PeerConnectionController {
...
-            this.peerConnection?.addTransceiver('video', { direction: 'recvonly' });
+
+        // Setup a transceiver for sending webcam video to UE and receiving video from UE
+        if (!useCamera) {
+            if(!hasVideoReceiver) {
+                this.peerConnection?.addTransceiver('video', { direction: 'recvonly' });
+            }
+        } else {
+            // set the media send options
+            const mediaSendOptions: MediaStreamConstraints = {
```
lukehb: Can the else block be moved out into its own function like `setupVideoSender()`?

lukehb (changes requested):
Minor change requesting transceiver setup with camera to be factored out into its own function.

lukehb:
Prettier is also catching some issues with the code it seems.

mcottontensor:
Looks like indenting issues. Can probably just use `npm run lint -- --fix`

PR #253, opened 2024-08-27
Adding prettier to the linter rules.

mcottontensor:
Also ran prettier over all the frontend source. Should be less irritating than clang-format.


lukehb (approved):
Happy with it!

PR #260, opened 2024-08-29
Updating keypress to use helper to deal with charCode deprecation.

mcottontensor:
Just to bring it in line with down and up event handling.


lukehb (approved):
lgtm

PR #266, opened 2024-09-02
JS Streamer and Frontend Tests

mcottontensor:
JS Streamer in Extras/JSStreamer
Frontend tests in Extras/FrontendTests


Extras/FrontendTests/tests/extras.ts
```diff
@@ -0,0 +1,86 @@
...
+    command: string;
+};
+
+// a generic type for inputs captured by the streamer
+export type DataChannelEvent = DataChannelMouseInput | DataChannelKeyboardInput | DataChannelCommandInput;
+
+// sets up the streamer page to capture data channel messages
+export function setupEventCapture(streamerPage: Page) {
+    return streamerPage.evaluate(() => {
+        window.data_messages = {};
```
lukehb: Maybe small comment here about attaching this js object and the listener to populate it to the window as (I imagine) this is the most convenient place that is visible to both caller and callee.

Extras/FrontendTests/tests/helpers.ts
```diff
@@ -0,0 +1,130 @@
...
+        const video_aspect_ratio = this.video_size.height / this.video_size.width;
+        this.player_is_larger = player_aspect_ratio > video_aspect_ratio;
+        if (this.player_is_larger) {
+            this.ratio = player_aspect_ratio / video_aspect_ratio;
+        } else {
+            this.ratio = video_aspect_ratio / player_aspect_ratio;
+        }
+    }
+
+    public normalizeQuantizeSigned(x: number, y: number): Coord {
```
lukehb, line 95: Could be potential to move somethings into common here and have both frontend and extra use it?

Extras/FrontendTests/tests/keyboard.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+
+test('Test keyboard events', {
+    tag: ['@keyboard'],
+}, async ({ page, streamerPage, streamerId }) => {
+
+    await page.goto(`/?StreamerId=${streamerId}&MatchViewportRes=true`);
+    await page.getByText('Click to start').click();
+    await helpers.waitForVideo(page);
+
+    const player_box = await page.locator('#videoElementParent').boundingBox();
```
lukehb: Any particular reason for `underscored_variables` - seems a departure from the rest of the codebase?

Extras/FrontendTests/tests/keyboard.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+                expected_actions.push({ type: PSEventTypes.KeyDown, key_code: c });
+                expected_actions.push({ type: PSEventTypes.KeyPress, key_code: c });
+                expected_actions.push({ type: PSEventTypes.KeyUp, key_code: c });
+            }
+        });
+
+        // check we got the expected events
+        const first_player_id = Object.keys(events)[0];
+        const single_player_events = events[first_player_id];
+        expect(single_player_events).toContainActions(expected_actions);
```
lukehb: That's nice!

Extras/FrontendTests/tests/helpers.ts
mcottontensor: Yeah I was thinking the same thing as I was doing this. I feel like the input stuff on frontend could be cleaned up a lot.

Extras/FrontendTests/tests/keyboard.spec.ts
mcottontensor: It was picked up from somewhere. Maybe when I started doing playwright tests, and was trying to be consistent and it became a habit when moving into this code base.
The joys of trying to be consistent with everyone.

lukehb (approved):

PR #275, opened 2024-09-04
Cleaning up a bunch of the input code.

mcottontensor:
Sorting out the classes under Input in the frontend library to be a little easier to understand.


Frontend/library/src/Inputs/GamepadController.ts
```diff
@@ -1,59 +1,82 @@
...
 
-    /**
-     * @param toStreamerMessagesProvider - Stream message instance
-     */
-    constructor(toStreamerMessagesProvider: StreamMessageController) {
-        this.toStreamerMessagesProvider = toStreamerMessagesProvider;
+    constructor(streamMessageController: StreamMessageController) {
+        this.streamMessageController = streamMessageController;
 
+        this.onGamepadConnectedListener = this.gamePadConnectHandler.bind(this);
```
lukehb: This class is a shocker for casing.

`gamePad`, `gamepad`, and `Gamepad` all seen here

Maybe `gamePad` ones can be fixed up. Those seem to be the worst abuse.

Frontend/library/src/Inputs/GamepadController.ts
```diff
@@ -65,31 +88,29 @@ export class GamePadController {
         }
     }
```
lukehb, line 95: Just for sanity, can you confirm streaming and using a gamepad to move around etc is still working?

Frontend/library/src/Inputs/MouseController.ts
```diff
@@ -253,56 +120,122 @@ export class MouseController {
...
+        this.streamMessageController.toStreamerHandlers.get('MouseDown')?.([button, X, Y]);
+    }
+
+    private sendMouseUp(button: number, X: number, Y: number) {
+        const coord = this.coordinateConverter.normalizeAndQuantizeUnsigned(X, Y);
+        this.streamMessageController.toStreamerHandlers.get('MouseUp')?.([button, coord.x, coord.y]);
+    }
+}
+
+export class HoveringMouseController extends MouseController {
```
lukehb: HoverMouseController and MouseControllerHovering?

Is this a copy/paste error?

Frontend/library/src/Inputs/MouseController.ts
```diff
@@ -1,212 +1,85 @@
...
+    streamMessageController: StreamMessageController;
     coordinateConverter: CoordinateConverter;
-    activeKeysProvider: ActiveKeys;
+    activeKeys: ActiveKeys;
 
-    // Utility for keeping track of event handlers and to unregister them.
-    private mouseEventListenerTracker = new EventListenerTracker();
+    // bound listeners
+    onEnterListener: (event: MouseEvent) => void;
+    onLeaveListener: (event: MouseEvent) => void;
```
lukehb, line 34: These always struck me as odd. Who is the caller?

Frontend/library/src/Inputs/MouseController.ts
```diff
@@ -253,56 +120,122 @@ export class MouseController {
...
-        toStreamerHandlers.get('MouseDown')([button, X, Y]);
+        const coord = this.coordinateConverter.normalizeAndQuantizeUnsigned(event.offsetX, event.offsetY);
+        const delta = this.coordinateConverter.normalizeAndQuantizeSigned(event.movementX, event.movementY);
+        this.streamMessageController.toStreamerHandlers.get('MouseMove')([
+            coord.x,
+            coord.y,
+            delta.x,
+            delta.y
+        ]);
+        event.preventDefault();
```
lukehb: I question preventDefault() on mousemove (I understand this is directly ported).

My reason for question it is because preventDefault() will cause the next event to be delayed by a pretty decent ms (like 50ms I have read).

Frontend/library/src/Inputs/MouseController.ts
```diff
@@ -253,56 +120,122 @@ export class MouseController {
...
-        const coord = this.coordinateConverter.normalizeAndQuantizeUnsigned(X, Y);
-        const toStreamerHandlers = this.toStreamerMessagesProvider.toStreamerHandlers;
-        toStreamerHandlers.get('MouseUp')([button, coord.x, coord.y]);
+        const coord = this.coordinateConverter.normalizeAndQuantizeUnsigned(event.offsetX, event.offsetY);
+        this.streamMessageController.toStreamerHandlers.get('MouseWheel')([
+            event.wheelDelta,
+            coord.x,
+            coord.y
+        ]);
+        event.preventDefault();
```
lukehb: This one I can understand, to prevent page scrolling

Frontend/library/src/Inputs/MouseControllerHovering.ts
```diff
@@ -0,0 +1,115 @@
...
+        }
+        const coord = this.coordinateConverter.normalizeAndQuantizeUnsigned(event.offsetX, event.offsetY);
+        const delta = this.coordinateConverter.normalizeAndQuantizeSigned(event.movementX, event.movementY);
+        this.streamMessageController.toStreamerHandlers.get('MouseMove')([
+            coord.x,
+            coord.y,
+            delta.x,
+            delta.y
+        ]);
+        event.preventDefault();
```
lukehb, line 95: Same here about preventDefault()

Frontend/library/src/Inputs/ITouchController.ts
```diff
@@ -4,26 +4,5 @@
...
-    /**
-     * Fire when touch moves
-     * @param touchEvent - Touch Event Data
-     */
-    onTouchMove(touchEvent: TouchEvent): void;
-
-    /**
-     * Unregister all touch event handlers.
-     */
     unregisterTouchEvents(): void;
```
lukehb: Is this interface even really that useful anymore? It only has one unregister method?

Frontend/library/src/Inputs/MouseControllerLocked.ts
```diff
@@ -0,0 +1,192 @@
+import { Logger } from '@epicgames-ps/lib-pixelstreamingcommon-ue5.5';
+import { StreamMessageController } from '../UeInstanceMessage/StreamMessageController';
+import { CoordinateConverter } from '../Util/CoordinateConverter';
+import { VideoPlayer } from '../VideoPlayer/VideoPlayer';
+import type { ActiveKeys } from './InputClassesFactory';
+import { NormalizedQuantizedUnsignedCoord } from '../Util/CoordinateConverter';
+import { MouseController } from './MouseController';
+
+export class MouseControllerLocked extends MouseController {
```
lukehb, line 11: Docstring?

Frontend/library/src/Util/CoordinateConverter.ts
```diff
@@ -1,218 +1,73 @@
...
-        };
+    playerIsLarger: boolean;
+
+    // we dont use a constructor here because the object is created and passed around to various locations
+    // possibly before this method is called.
+    configure(playerSize: RectSize, videoSize: RectSize) {
+        const playerAspectRatio = playerSize.height / playerSize.width;
+        const videoAspectRatio = videoSize.height / videoSize.width;
+        this.playerIsLarger = playerAspectRatio > videoAspectRatio;
+        this.playerSize = playerSize;
```
lukehb: Caution about passing in a playerSize and videoSize like this, I think they can change as the page is resized, hence why the element was used before.

Frontend/library/src/WebRtcPlayer/WebRtcPlayerController.ts
```diff
@@ -1386,7 +1386,12 @@ export class WebRtcPlayerController {
     setUpMouseAndFreezeFrame() {
         // Calculating and normalizing positions depends on the width and height of the player.
         this.videoElementParentClientRect = this.videoPlayer.getVideoParentElement().getBoundingClientRect();
-        this.coordinateConverter.setupNormalizeAndQuantize();
+        const playerElement = this.videoPlayer.getVideoParentElement();
+        const videoElement = this.videoPlayer.getVideoElement();
+        this.coordinateConverter.configure(
+            { width: playerElement.clientWidth, height: playerElement.clientHeight },
+            { width: videoElement.videoWidth, height: videoElement.videoHeight }
+        );
```
lukehb, line 1390: I think if you were gonna do it like this you also have to bind to element resize events to call configure then too.

lukehb: Sanity check will be testing with page resizing and mouse coordinates still working as expected (probably more obvious when UE cursor is showing)

lukehb (changes requested):
Left some comments.

Frontend/library/src/Util/CoordinateConverter.ts
mcottontensor: Configure is called from an onresize event, Thats why i called it `configure` instead of `init` or `setup` or similar.

lukehb:
Oh, guess I missed that.

Frontend/library/src/Inputs/MouseController.ts
mcottontensor: I'm not sure what you mean? These are for binding callbacks to `this` and saving the result so we can remove them later on.

Frontend/library/src/WebRtcPlayer/WebRtcPlayerController.ts
mcottontensor: This is not really any different from how it was. I just moved the element holding out of the util class so it could be a little more generic. The way it grabs and calculates sizes is still the same.

Frontend/library/src/Inputs/MouseControllerHovering.ts
mcottontensor: I can remove them but I didn't want to alter the existing behaviour too much.

lukehb: Let's keep it for now.

PR #247, opened 2024-08-21
Add support for configurable countdown value for AFKTimeout

pr0g:
## Relevant components:
- [ ] Signalling server
- [ ] Common library
- [X] Frontend library
- [X] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:

As a user of the **PixelStreamingInfrastructure** library, I would like to be able to configure the countdown provided to users after an AFK timeout has elapsed. Currently this countdown value is hardcoded to 10 seconds.

## Solution

This PR adds a new `NumericParameter` called `AFKCountdown` which can be configured at the same time as other AFK properties when the `PixelStream` is first created. The default value is 10 seconds, but this can be set to be a total of 3 minutes if desired (this upper limit can totally be adjusted based on PR feeedback, the duration we'd like to use on our team is 60 seconds).

## Documentation

I have updated `Frontend/Docs/Timing Out Inactive Connections.md` to include a reference to the new variable/setting (I also fixed the formatting of the Markdown table which wasn't rendering correctly). 

## Test Plan and Compatibility

I have not yet created tests to verify the feature yet (if this is something required for the PR I wanted to wait to get the feedback/comments first).

I did find existing tests in `AFKController.test.ts` which exercise the countdown feature (see test `'should activate AFK timer and trigger it after specified delay if it has been enabled from settings'`). I'd be happy to update this test to verify a different number of seconds for the countdown if that would be useful.

## Context

We initially created this change off of the [UE5.3-1.0.4](https://github.com/revu-design/PixelStreamingInfrastructure/releases/tag/UE5.3-1.0.4) tag that we're using for our project. The changes have been rebased onto the master branch for this PR.

PR #287, opened 2024-09-27
Update max value for AFK countdown and timeout

pr0g:
## Relevant components:
- [ ] Signalling server
- [ ] Common library
- [X] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:

This addresses an issue we've encountered where the AFK countdown and timeout delays are still too short.

## Solution

This PR bumps the max value supported for both AFK countdown and timeout.

## Documentation

N/A for this PR, the min/max values are not documented and documentation was added in the previous PR (see https://github.com/EpicGamesExt/PixelStreamingInfrastructure/pull/247).

We plan to internally test this with our fork to ensure we can set longer timeouts.

## Additional

We would really appreciate if this change could be backported to `UE5.3-1` (similar to this release - https://github.com/EpicGamesExt/PixelStreamingInfrastructure/releases/tag/UE5.3-1.0.6). The diff for that version is slightly different (it's actually simpler) and looks like this:

```diff
diff --git a/Frontend/library/src/Config/Config.ts b/Frontend/library/src/Config/Config.ts
index cd02bb9..7402e6f 100644
--- a/Frontend/library/src/Config/Config.ts
+++ b/Frontend/library/src/Config/Config.ts
@@ -472,7 +472,7 @@ export class Config {
                 'AFK timeout',
                 'The time (in seconds) it takes for the application to time out if AFK timeout is enabled.',
                 0 /*min*/,
-                600 /*max*/,
+                3600 /*max*/,
                 120 /*value*/,
                 useUrlParams
             )
@@ -485,7 +485,7 @@ export class Config {
                 'AFK countdown',
                 'The time (in seconds) for a user to respond before the stream is ended after an AFK timeout.',
                 10 /*min*/,
-                180 /*max*/,
+                600 /*max*/,
                 10 /*value*/,
                 useUrlParams
             )
```

### DIFF
--- a/Frontend/library/src/Config/Config.ts
+++ b/Frontend/library/src/Config/Config.ts
@@ -506,10 +506,10 @@ export class Config {
                 'AFK timeout',
                 'The time (in seconds) it takes for the application to time out if AFK timeout is enabled.',
                 0 /*min*/,
-                600 /*max*/,
+                3600 /*max - 1 hour*/,
                 settings && Object.prototype.hasOwnProperty.call(settings, NumericParameters.AFKTimeoutSecs)
                     ? settings[NumericParameters.AFKTimeoutSecs]
-                    : 120 /*value*/,
+                    : 120 /*default value - 2 minutes*/,
                 useUrlParams
             )
         );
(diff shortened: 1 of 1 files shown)


mcottontensor:
This looks alright to me but I have a slightly different proposal.

I put up a new PR #288 which adds the ability to provide undefined for the min and max values and removes the limit for the AFK timers. So now users can set whatever value they need without limit.

Let me know if this works for you and ill back port it to the supported branches.

pr0g:
Hi @mcottontensor, sorry I'm just getting back to you. 

Your proposal is a great idea! I think the PR looks excellent. My only minor suggestion would be it might make more sense to use `null` instead of `undefined` in this case as it implies an intentionally empty value instead of a missing one. It's a small semantic difference but might make the code slightly clearer.

Feel free to ignore my suggestion, the overall change looks great and it would be fantastic if you could backport it too.

Thank you very much for your help.

mcottontensor:
Have merged #288 to master, UE5.2, UE5.3 and UE5.4 and pushed new releases for each branch. This should cover your case. Cheers.

pr0g:
That's brilliant, @mcottontensor, thank you very much! 🙂

PR #299, opened 2024-10-13
Fix unhandled error on client from SettingOption.ts

richardFocus:
## Relevant components:
- [ ] Signalling server
- [ ] Common library
- [X] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
I have been having unhandled errors in production and found this update solves the issue we were having originating from SettingOptions.ts:
settingOptionsError

## Solution
The was an issue with the options not being ready in time, I am not sure if this is the best way to solve the bug as it may be better to solve elsewhere in the codebase but this has fixed the issue for us and now I don't get anymore production errors, if someone else knows a better fix then go for it.

## Testing
I have been running this code in productions for weeks now and no more issues that were occurring previously at least every few days.

### DIFF
--- a/Frontend/library/src/Config/SettingOption.ts
+++ b/Frontend/library/src/Config/SettingOption.ts
@@ -76,7 +76,12 @@ export class SettingOption<CustomIds extends string = OptionParametersIds> exten
         // A user may not specify the full possible value so we instead use the closest match.
         // eg ?xxx=H264 would select 'H264 level-asymmetry-allowed=1;packetization-mode=1;profile-level-id=42001f'
         if (!value || typeof value !== 'string' || value.trim() === '') {
-            Logger.Warning(`Invalid argument type for the selected codec, argument type: ${typeof value}, with value: ${value}`);
+            Logger.Warning(
+                `
+                Invalid argument type for the selected codec,
+                argument type: ${typeof value}, with value: ${value}
+                `
+            );
             return;
         }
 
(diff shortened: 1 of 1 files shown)


lukehb (changes requested):
The change to add some error handling seem okay.

@mcottontensor I think the logging to console should use our logging class instead, right?

Frontend/library/src/Config/SettingOption.ts
```diff
@@ -74,18 +74,32 @@ export class SettingOption<CustomIds extends string = OptionParametersIds> exten
...
             this.value = filteredList[0];
             return;
         }
 
-        // A user has specified a codec with a fmtp string but this codec + fmtp line isn't available.
-        // in that case, just use the codec
-        filteredList = this.options.filter((option: string) => option.indexOf(value.split(' ')[0]) !== -1);
-        if (filteredList.length) {
+        // If no match is found, and value contains parameters, try matching only the codec
+        const valueCodec = value.split(' ')[0];
```
lukehb, line 98: `value.split(' ')[0]` could itself fail (same with above on line 83) - I would request some defensive programming be done around that too, e.g. check if split results in a non empty array, then that value [0] is non-empty.

lukehb (changes requested):
Some further changes come to mind making it even safer.

Frontend/library/src/Config/SettingOption.ts
richardFocus: Yes you are correct.

richardFocus (commented):
Updated as requested.

richardFocus (commented):
Updated import path for Logger function

richardFocus (commented):
Removed getStrackTrace()

richardFocus (commented):
linting fix

richardFocus (commented):
fixed linting

lukehb:
Thanks for your contribution @richardFocus - I ended up having to change a bunch of stuff related to this because of some other work I was doing - you can see the landed result in: https://github.com/EpicGamesExt/PixelStreamingInfrastructure/pull/324

PR #300, opened 2024-10-13
Bringing back the functionality that prevents unidentified streamers showing up in the streamer list

mcottontensor:
## Relevant components:
- [x] Signalling server
- [x] Common library
- [ ] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
Streamers can show up in the streamer list before they've identified and started streaming.

## Solution
Added a filter to filter out streamers that aren't streaming.

Additionally regenerated the proto files.

### DIFF
--- a/Signalling/src/PlayerConnection.ts
+++ b/Signalling/src/PlayerConnection.ts
@@ -116,6 +116,10 @@ export class PlayerConnection implements IPlayer, LogUtils.IMessageLogger {
         this.protocol.on(Messages.dataChannelRequest.typeName, this.sendToStreamer.bind(this));
         this.protocol.on(Messages.peerDataChannelsReady.typeName, this.sendToStreamer.bind(this));
         this.protocol.on(Messages.layerPreference.typeName, this.sendToStreamer.bind(this));
+
+        this.protocol.on('unhandled', (message: BaseMessage) => {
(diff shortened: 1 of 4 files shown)


Signalling/src/PlayerConnection.ts
```diff
@@ -210,7 +214,9 @@ export class PlayerConnection implements IPlayer, LogUtils.IMessageLogger {
 
     private onListStreamers(_message: Messages.listStreamers): void {
         const listMessage = MessageHelpers.createMessage(Messages.streamerList, {
-            ids: this.server.streamerRegistry.streamers.map((streamer) => streamer.streamerId)
+            ids: this.server.streamerRegistry.streamers
+                .filter((streamer) => streamer.streaming)
```
lukehb, line 218: How does a streamer become `streaming` again? Just sanity checking this.

mcottontensor: When the streamer responds to the endpoint id request its marked as streaming.

lukehb: Okay, I think that works well, we just will ensure we don't backport this one.

lukehb (approved):

PR #323, opened 2024-11-21
fix: change typo in unregister callbacks to remove event listeners

lucastanger:
## Relevant components:
- [ ] Signalling server
- [ ] Common library
- [x] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
This PR targets the TouchControllers unregister function, which re-added the same event listeners instead of removing them.

## Solution
Changed from addEventListener to removeEventListener


lukehb (approved):
Thanks for catching this one. 

CC @mcottontensor

PR #325, opened 2024-11-24
Bring changes from 5.5 into master

lukehb:


lukehb:
Publish actions were expected to fail as I already pushed these manually.

Issue #328, opened 2024-11-25
[BUG] - @epicgames-ps/lib-pixelstreamingfrontend-ue5.5 is including EventEmitter, which causes build problems


leon:
**UE Version:**
E.g. UE 5.5.0

**Frontend Version:**
E.g. UE5.5-0.3.2

**Problem component**
E.g. Frontend

**Description**
I have a very simple vite project that is importing
```typescript
import {
	Config,
	PixelStreaming,
} from '@epicgames-ps/lib-pixelstreamingfrontend-ue5.5';
```

I tried upgrading from v 2.1.0 to 0.3.2 but I'm now getting Module "events" has been externalized for browser compatibility. Cannot access "events.EventEmitter"...

The problem seems to be that `@epicgames-ps+lib-pixelstreamingcommon-ue5.5` `WebSocketTransport.js` is getting included in the frontend code and it uses nodes EventEmitter, which is a node only library.

**Steps to Reproduce:**
create a new vite project
use v0.3.2
try building

here is a repo that you can test out
https://github.com/leon/bug-pixelstreaming-eventemitter

**Expected behavior**
Should be able to build frontend using 5.5 v0.3.2

**Screenshots**

**Platform (please complete the following information):**
MacOS Sequoia

lukehb:
Thanks for reporting, I will look into this today.

lukehb:
Working on a PR - should have it fixed tomorrow: https://github.com/EpicGamesExt/PixelStreamingInfrastructure/pull/330

lukehb:
If you try bumping to the latest (0.4.0) @epicgames-ps/lib-pixelstreamingfrontend-ue5.5 it should be fixed for you (I tested against your provided repo and it built).

leon:
Hey,

Just wanted to mention that 0.4.0 is not tagged as the latest release on npm
https://www.npmjs.com/package/@epicgames-ps/lib-pixelstreamingfrontend-ue5.5?activeTab=versions

mcottontensor:
Yeah thats because I had to push a fix for the 0.3.x version.
@lukehb You might be able to make it as latest, but I don't have direct perms.

PR #330, opened 2024-11-26
Removed NodeJS event emitter in common library.

lukehb:
## Relevant components:
- [x] Signalling server
- [x] Common library
- [x] Frontend library
- [x] Frontend UI library

## Problem statement:
The `Common` library contained a NodeJS type, `EventEmitter` this was causing it to fail when used in the browser. This was not a problem previously when we bundled at the library build step, but since https://github.com/EpicGamesExt/PixelStreamingInfrastructure/pull/324 have been a few dependencies that do not work well in the browser without bundling.

This issue was reported in: https://github.com/EpicGamesExt/PixelStreamingInfrastructure/issues/328 and impacts the latest version of UE 5.5 frontend library.

## Solution
This PR solves the issue by making a browser compatible EventEmitter type using EventTarget under the hood.

## Documentation
N/A

I have tested by:
- Running the unit tests
- Running the frontend tests
- Running with an actual UE instance
- Running the signalling server


Common/src/Event/EventEmitter.ts
```diff
@@ -0,0 +1,277 @@
...
+
+class PixelStreamingEventListener implements EventListenerObject {
+    private _callback: Function;
+    private _args: any[] = [];
+
+    constructor(callback: Function) {
+        this._callback = callback;
+    }
+
+    handleEvent(_evt: Event): void {
```
mcottontensor, line 17: Should the event not be passed to the callback?

Common/src/Event/EventEmitter.ts
```diff
@@ -0,0 +1,277 @@
...
+        this._callback(...this._args);
+    }
+
+    set args(value: any) {
+        this._args = value;
```
mcottontensor: I'm guessing when you assign an any[] parameter to this.args it comes through as any[] which is also covered by any.. Should this be more explicit?

Extras/FrontendTests/tests/mouse.spec.ts
```diff
@@ -108,7 +114,7 @@ test('Test locked mouse movement', {
         console.log("Player: ", ...args);
     });
 
-    await page.goto(`/?StreamerId=${streamerId}&MatchViewportRes=true&HoveringMouse=false`);
+    await page.goto(`/?StreamerId=${streamerId}&MatchViewportRes=false&HoveringMouse=false`);
```
lukehb:
```suggestion
    await page.goto(`/?StreamerId=${streamerId}&MatchViewportRes=true&HoveringMouse=false`);
```

Extras/FrontendTests/tests/mouse.spec.ts
```diff
@@ -238,7 +244,7 @@ test('Test mouse input after resizing. Hover mouse.', {
     //     console.log("Player: ", ...args);
     // });
 
-    await page.goto(`/?StreamerId=${streamerId}&MatchViewportRes=true&HoveringMouse=true`);
+    await page.goto(`/?StreamerId=${streamerId}&HoveringMouse=true`);
```
lukehb:
```suggestion
    await page.goto(`/?StreamerId=${streamerId}&MatchViewportRes=true&HoveringMouse=true`);
```

Extras/FrontendTests/tests/mouse.spec.ts
```diff
@@ -296,7 +308,7 @@ test('Test mouse input after resizing. locked mouse.', {
     //     console.log("Player: ", ...args);
     // });
 
-    await page.goto(`/?StreamerId=${streamerId}&MatchViewportRes=true&HoveringMouse=false`);
+    await page.goto(`/?StreamerId=${streamerId}&HoveringMouse=false`);
```
lukehb:
```suggestion
    await page.goto(`/?StreamerId=${streamerId}&MatchViewportRes=true&HoveringMouse=false`);
```

Extras/JSStreamer/package.json
```diff
@@ -7,7 +7,7 @@
   "module": "build/pixelstreamingjsstreamer.js",
   "scripts": {
     "develop": "webpack-dev-server --mode development",
-    "build": "tsc --project tsconfig.json && webpack --mode production",
+    "build": "tsc --project tsconfig.json && webpack --mode development",
```
lukehb, line 10: Didn't really make sense to minify/obscure this code as it is only used in dev.

Extras/JSStreamer/webpack.config.js
```diff
@@ -45,7 +45,7 @@ module.exports = {
             },
         },
         static: path.join(__dirname, 'dist'),
-        compress: true,
+        compress: false,
```
lukehb, line 48: Didn't make sense to compress this code as we only use it for testing/dev.

Frontend/library/src/Util/EventEmitter.ts
```diff
@@ -581,7 +581,7 @@ export type PixelStreamingEvent =
     | PlayerCountEvent
     | WebRtcTCPRelayDetectedEvent;
 
-export class EventEmitter extends EventTarget {
+export class PixelStreamingEventEmitter extends EventTarget {
```
lukehb, line 584: We already had a class called EventEmitter, so I renamed it here. This seemed like the lesser impactful rename as opposed to renaming the one in common.

Extras/FrontendTests/package.json
```diff
@@ -4,13 +4,14 @@
   "description": "",
   "main": "index.js",
   "scripts": {
-    "test": "npx playwright test"
+    "test": "npx playwright test",
+    "build": ""
```
lukehb, line 8: You may notice a few empty build scripts in this commit.

Those were added to enable running `npm run build` in the root of the repo and building all workspaces - handy.

SignallingWebServer/Dockerfile
```diff
@@ -1,6 +1,7 @@
 # Use the current Long Term Support (LTS) version of Node.js
 FROM node:lts
 # Copy the signalling server and frontend source code from the build context
+COPY package.json package.json
```
lukehb, line 4: This was needed in the Dockerfile to make it use the new NPM workspaces way of doing things.

package.json
```diff
@@ -15,5 +15,8 @@
 		"SignallingWebServer",
 		"SS_Test"
 	],
-	"private": true
+	"private": true,
+	"scripts": {
+		"build": "npm run build --ws"
+	}
```
lukehb, line 21: This lets us run `npm run build` in the repo root and build everything.

Common/src/Event/EventEmitter.ts
mcottontensor: Still questioning this one.

mcottontensor (approved):

Common/src/Event/EventEmitter.ts
lukehb: passing that Event to the downstream listener would actually introduce different behaviour than what node js was doing with its event emitter

It has no meaningful information attached to it and is really an internal implementation detail, exposing it to the listener would be a leaky abstraction considering the point was the make something that looks and smells likes NodeJS event emitter but actually uses EventTarget under the hood.

Common/src/Event/EventEmitter.ts
lukehb: Has been updated now.

Issue #234, opened 2024-08-08
Make frontend changes to accomodate min/max encoder quality changes in 5.5


DenisTensorWorks:
**Component your feature request relates to**
- [x] Frontend library
- [x] Frontend UI library

**Problem your feature solves? Please describe.**
We need to accomodate for the move to min/max encoder quality scale instead of min/max QP cvars. 
Ensure we still send the old MinQP/MaxQP messages for backwards compatibility.

Internal reference: UE-220968

lukehb:
@mcottontensor When you get round to this, let's have a discussion. But basically some internal discussion of changing these on the plugin side, which has some flow on effect to the frontend.

PR #345, opened 2024-12-03
Adding the compatibility quality min/max

mcottontensor:
## Relevant components:
- [ ] Signalling server
- [ ] Common library
- [x] Frontend library
- [x] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
We want to simplify the quality values exposed to the user to be pulled into line with PS2

## Solution
A new middle man config value called CompatQualityMin/Max that exposes values of 0-100 to the user but behind the scenes converts to 0-51 to send to the streamer. This is to allow PS1 to use it and PS2 to convert it to quality.
In order to support URL parameters that directly modify either QP or Quality values, the existing QP value pairs still remain and another pair for Quality have been added that will send the values directly to the streamer and its left up to the streamer to interpret. These values work with the new CompatQuality values to reflect the correct quality level in the UI.

#bypass-publish-check

### DIFF
--- a/Frontend/library/src/WebRtcPlayer/WebRtcPlayerController.ts
+++ b/Frontend/library/src/WebRtcPlayer/WebRtcPlayerController.ts
@@ -1488,6 +1488,44 @@ export class WebRtcPlayerController {
     }
 
     /**
+     * Send the MinQuality encoder setting to the UE Instance.
+     * @param minQuality - The lower bound for quality when encoding
+     * valid values are (0-100) where:
+     * 0 = Worst quality.
+     * 100 = Best quality.
+     */
+    sendEncoderMinQuality(minQuality: number) {
+        Logger.Info(`MinQuality=${minQuality}\n`);
+
+        if (minQuality != null) {
+            this.streamMessageController.toStreamerHandlers.get('Command')([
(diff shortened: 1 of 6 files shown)


Frontend/library/src/WebRtcPlayer/WebRtcPlayerController.ts
```diff
@@ -1487,6 +1487,44 @@ export class WebRtcPlayerController {
...
+            ]);
+        }
+    }
+
+    /**
+     * Send the MaxQuality encoder setting to the UE Instance.
+     * @param maxQuality - The upper bound for quality when encoding
+     * valid values are (0-100) where:
+     * 0 = Best quality.
+     * 100 = Worst quality.
```
aussiedwarf: Shouldn't this and one above be 0 = Worst quality, 100 = Best quality.'

mcottontensor: Probably

lukehb (approved):

Belchy06 (approved):

mcottontensor:
Closes #234

PR #347, opened 2024-12-06
Adding git tags to each publish action.

mcottontensor:
## Relevant components:
- [ ] Signalling server
- [ ] Common library
- [ ] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU
- [x] Other

Each component that is published will now use its package.json file/version number to create a tag at the point of publishing.

This should give us a nice foundation for Github releases of these packages with changelog generation in the future.

### DIFF
--- a/.github/workflows/publish-common-library-to-npm.yml
+++ b/.github/workflows/publish-common-library-to-npm.yml
@@ -2,7 +2,7 @@ name: Publish common lib
 on:
   workflow_dispatch:
   push:
-    branches: ['UE5.5']
+    # branches: ['UE5.5']
     paths: ['Common/package.json']
 
 env:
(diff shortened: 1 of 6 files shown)


.github/workflows/publish-common-library-to-npm.yml
```diff
@@ -20,6 +20,24 @@ jobs:
...
+          echo $package_content >> $GITHUB_OUTPUT
+          echo 'EOF' >> $GITHUB_OUTPUT
+
+      - uses: actions/github-script@v5
+        with:
+          script: |
+            github.rest.git.createRef({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              ref: 'refs/tags/lib-common-v${{ fromJson(steps.get-package-json.outputs.json).version }}',
```
lukehb: Can you remove the `v` from all of these?

It is not valid semver, which means the GH auto releases won't find the right range.

https://semver.org/#is-v123-a-semantic-version

lukehb (changes requested):

lukehb (approved):

PR #350, opened 2024-12-09
Conditional release

mcottontensor:
## Relevant components:
- [ ] Signalling server
- [ ] Common library
- [ ] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU
- [x] Other

Updated the release action to trigger the check of public npm libraries. If the public libraries don't work with the release the release action will fail.

When giving a commitsha the library check action should also check against that sha.

### DIFF
--- a/.github/workflows/create-gh-release.yml
+++ b/.github/workflows/create-gh-release.yml
@@ -15,20 +15,24 @@ env:
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


lukehb (approved):
lgtm

PR #356, opened 2024-12-10
Add: Preferred quality option (and UI) for simulcast or svc enabled streams

Belchy06:
## Relevant components:
- [ ] Signalling server
- [X] Common library
- [X] Frontend library
- [X] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [X] SFU

## Problem statement:
Unreal Engine 5.5 removed the ability to select a preferred SFU layer with blueprints meaning that currently, there is no way to indicate the preferred layer to the SFU. 

## Solution
This PR moves the preferred layer selection to the frontend with easily readable names like "Low", "Medium" or "High" when using simulcast or if using VP9 SVC, the dropdown will show all possible spatio-temporal combinations. Some additional QOL improvements include:
- Ability to programmatically enable or disable a UI setting post initialization. This is used to disable the quality dropdown if connecting to a default stream or enabling if connecting to the SFU.
- Moved the logging of a forwarded message in the signalling server to until after the `playerId` field has been set for easier debugging.

## Documentation
Updated docs to reflect the new setting.


Frontend/ui-library/tsconfig.json
```diff
@@ -4,7 +4,7 @@
         "noImplicitAny": true,
         "module": "es6",
         "esModuleInterop": true,
-        "target": "ES6",
+        "target": "ES2015",
```
lukehb: Why is this changed throughout?

lukehb (dismissed):

Frontend/ui-library/tsconfig.json
Belchy06: I was getting compiler warnings in `Config.ts` that it couldn't compile unless the target was set to ES2015 or newer.

mcottontensor (approved):

PR #370, opened 2024-12-17
fix byte incosistency

lucastanger:
## Relevant components:
- [ ] Signalling server
- [ ] Common library
- [x] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
The pre-defined 16kB were defined as 16379, which is not correct.

## Solution
Either use the real number of 16kB (16384) or compute it on the fly with 16 * 1024 (as done below when calculating the transferBitrate)

I recently implemented a screenshot feature in our UI that requests an image from the engine. This image is transferred using your "FreezeFrame" functionality. We can adjust our request by passing a desired width and height, which can sometimes result in missing bytes due to an incorrect calculation of the chunks to receive.

### DIFF
--- a/Frontend/library/src/Util/FileUtil.ts
+++ b/Frontend/library/src/Util/FileUtil.ts
@@ -61,7 +61,7 @@ export class FileUtil {
         // Extract the total size of the file (across all chunks)
         file.size = Math.ceil(
             new DataView(view.slice(1, 5).buffer).getInt32(0, true) /
-                16379 /* The maximum number of payload bits per message*/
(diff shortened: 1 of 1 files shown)


mcottontensor:
So the original code is actually correct but poorly written. The size variable is actually the count of chunks in the whole file. Each chunk has a few bytes for a header which include the message type and the file size. So the divisor here is actually the max message size minus the 5 bytes for the header data.

I have created an alternative PR #371 that should help the readability of the code. I have also tested it with a large file and confirmed it's correct.

PR #377, opened 2024-12-18
Add: Functionality to disable UI on a per section basis

Belchy06:
## Relevant components:
- [ ] Signalling server
- [ ] Common library
- [ ] Frontend library
- [X] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
Similar to https://github.com/EpicGamesExt/PixelStreamingInfrastructure/pull/372, but this time on a per section. This stems from the fact that you could disable all the settings in a section individually, but you would still get the section header remaining.

## Solution
Similar fashion to https://github.com/EpicGamesExt/PixelStreamingInfrastructure/pull/372

## Example Usage
Using the example below results in the attached image.
```ts
const application = new Application({
		stream,
		onColorModeChanged: (isLightMode) => PixelStreamingApplicationStyles.setColorMode(isLightMode),
		settingsPanelConfig: {
			isEnabled: true,
			settingVisibility: {
			  StreamerId: false,
			  ss: false,
			  TimeoutIfIdle: false,
			  ForceTURN: false,
			  LightMode: false,
			  UseMic: false,
			  MatchViewportRes: false,
			  HoveringMouse: false,
			  AFKTimeout: false,
			  AFKCountdown: false,
			  MaxReconnectAttempts: false,
			  StartVideoMuted: false
			},
			sectionVisibility: {
			  Encoder: false,
			  WebRTC: false,
			  UI: false,
			  Input: false
			}
		  }
	});
```

image

### DIFF
--- a/Frontend/implementations/typescript/src/player.ts
+++ b/Frontend/implementations/typescript/src/player.ts
@@ -22,7 +22,30 @@ document.body.onload = function() {
 
 	const application = new Application({
 		stream,
-		onColorModeChanged: (isLightMode) => PixelStreamingApplicationStyles.setColorMode(isLightMode)
+		onColorModeChanged: (isLightMode) => PixelStreamingApplicationStyles.setColorMode(isLightMode),
+		settingsPanelConfig: {
+			isEnabled: true,
+			settingVisibility: {
+			  StreamerId: false,
+			  ss: false,
+			  TimeoutIfIdle: false,
+			  ForceTURN: false,
+			  LightMode: false,
+			  UseMic: false,
+			  MatchViewportRes: false,
+			  HoveringMouse: false,
+			  AFKTimeout: false,
+			  AFKCountdown: false,
+			  MaxReconnectAttempts: false,
+			  StartVideoMuted: false
+			},
+			sectionVisibility: {
+			  Encoder: false,
+			  WebRTC: false,
+			  UI: false,
+			  Input: false
+			}
+		  }
 	});
 	document.body.appendChild(application.rootElement);
 
--- a/Frontend/implementations/typescript/src/player.ts
+++ b/Frontend/implementations/typescript/src/player.ts
@@ -22,30 +22,7 @@ document.body.onload = function() {
 
 	const application = new Application({
 		stream,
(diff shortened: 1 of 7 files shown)


Frontend/ui-library/src/Config/ConfigUI.ts
```diff
@@ -126,148 +128,160 @@ export class ConfigUI {
...
         }
 
-        if (isSettingEnabled(settingsConfig, OptionParameters.PreferredQuality))
-            this.addSettingOption(
-                encoderSettingsSection,
-                this.optionParametersUi.get(OptionParameters.PreferredQuality)
-            );
+        if (isSectionEnabled(settingsConfig, Sections.Encoder)) {
+            /* Setup all encoder related settings under this section */
+            const encoderSettingsSection = this.buildSectionWithHeading(settingsElem, 'Encoder');
```
lukehb: Perhaps heading should also come from the new `Sections` type instead of hardcoded string?

lukehb (changes requested):

Frontend/implementations/typescript/src/player.ts
```diff
@@ -22,7 +22,30 @@ document.body.onload = function() {
...
+			  MaxReconnectAttempts: false,
+			  StartVideoMuted: false
+			},
+			sectionVisibility: {
+			  Encoder: false,
+			  WebRTC: false,
+			  UI: false,
+			  Input: false
+			}
+		  }
```
lukehb: The typescript example shouldn't need any of this new config, right?

lukehb (changes requested):

Frontend/ui-library/src/Config/ConfigUI.ts
Belchy06: Yep, forgot to do that. Fixed!

Frontend/implementations/typescript/src/player.ts
Belchy06: Ah, bit of dev work seeping in. Fixed!

lukehb (approved):

lukehb (approved):

PR #379, opened 2024-12-19
Adding a pre commit check for npm libraries

mcottontensor:
## Relevant components:
- [ ] Signalling server
- [ ] Common library
- [ ] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU
- [x] Other

## Problem statement:
It's common to commit your work and build a PR only to later realise you forgot to run lint checks and the github actions fail, sometimes after a long while. This can lead to wasted time and frustrations.

## Solution
This installs a git pre-commit hook that will cycle through the npm libraries and if there are staged files in them it will run lint allowing you to get earlier notification of any linting issues before you actually commit. This should help reduce the number of "Fixing linting issues" commits.

This uses an npm package `pre-commit` in the root package.json which should automatically install the pre-commit hooks when you first do an `npm i`.

### DIFF
--- a/Extras/Scripts/pre-commit-checks.sh
+++ b/Extras/Scripts/pre-commit-checks.sh
@@ -1,0 +1,37 @@
+#!/bin/bash
+
+LINT_CHECK_PATHS=(
+    "Common/"
+    "Signalling/"
+    "Frontend/library/"
+    "Frontend/ui-library/"
+    "SignallingWebServer/"
+)
+
+FAILURE=0
+FAILED_PATHS=()
+
+for CHECK_PATH in "${LINT_CHECK_PATHS[@]}"; do
+    STAGED_FILES=$(git diff --name-only --cached "$CHECK_PATH")
+
+    # check if any of the npm packages have changes
+    if [[ -n "$STAGED_FILES" ]]; then
+        echo ${CHECK_PATH} has changes. Running lint...
+        pushd $CHECK_PATH >/dev/null
+        if ! npm run lint; then
+            FAILURE=1
+            FAILED_PATHS+=("$CHECK_PATH")
+        fi
+        popd >/dev/null
+    fi
+done
+
+if [[ $FAILURE -eq 1 ]]; then
+    echo Linting failures in the following paths...
+    for FAILED_PATH in "${FAILED_PATHS[@]}"; do
+        echo "    $FAILED_PATH"
+    done
+    exit 1
+fi
+
+
--- a/package.json
+++ b/package.json
@@ -17,6 +17,13 @@
 	],
 	"private": true,
 	"scripts": {
-		"build": "npm run build --ws"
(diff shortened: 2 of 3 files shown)


Belchy06 (approved):
LGTM

package.json
```diff
@@ -1,22 +1,29 @@
...
+        "SFU",
+        "SFU/mediasoup-sdp-bridge",
+        "Signalling",
+        "SignallingWebServer",
+        "SS_Test"
+    ],
+    "private": true,
+    "scripts": {
+        "build": "npm run build --ws",
+        "pre-commit-lint": "bash ./Extras/Scripts/pre-commit-checks.sh"
```
lukehb, line 21: What about Windows users? Won't this hard fail for them?

package.json
```diff
@@ -1,22 +1,29 @@
...
+    },
+    "pre-commit": [
+        "pre-commit-lint"
+    ],
+    "dependencies": {
```
lukehb, line 26: Should be dev-dependency, right?

Extras/Scripts/pre-commit-checks.sh
```diff
@@ -0,0 +1,37 @@
...
+
+if [[ $FAILURE -eq 1 ]]; then
+    echo Linting failures in the following paths...
+    for FAILED_PATH in "${FAILED_PATHS[@]}"; do
+        echo "    $FAILED_PATH"
+    done
+    exit 1
+fi
+
+
```
lukehb, line 37: Maybe this could be in JS?

```js
const { execSync } = require('child_process');

const LINT_CHECK_PATHS = [
    'Common/',
    'Signalling/',
    'Frontend/library/',
    'Frontend/ui-library/',
    'SignallingWebServer/'
];

let failure = false;
const failedPaths = [];

LINT_CHECK_PATHS.forEach((checkPath) => {
    try {
        // Check for staged files in the given path
        const stagedFiles = execSync(`git diff --name-only --cached ${checkPath}`).toString().trim();

        if (stagedFiles) {
            console.log(`${checkPath} has changes. Running lint...`);

            // Run lint command
            try {
                execSync('npm run lint', { cwd: checkPath, stdio: 'inherit' });
            } catch {
                failure = true;
                failedPaths.push(checkPath);
            }
        }
    } catch (err) {
        console.error(`Error checking path: ${checkPath}`, err);
    }
});

if (failure) {
    console.error('Linting failures in the following paths:');
    failedPaths.forEach((path) => console.error(`    ${path}`));
    process.exit(1);
}
```

lukehb: Nvm seems Windows git always comes with bash:

image

package.json
lukehb: When I just tried it, evidently not!

PR #398, opened 2025-01-14
Fix: ESM libraries not exporting types.

lukehb:
## Relevant components:
- [ ] Signalling server
- [x] Common library
- [x] Frontend library
- [x] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
ESM libraries were not exporting their types for use in browser using the following module import.

For example, consider:

```
<script type=module>
import * as PixelStreamingTypes from `./lib-pixelstreaming-UE5.5.js`;
console.log(PixelStreamingTypes); // This should print all the exported types/funcs
</script>
```

Currently without this PR, this does not log any types/functions as they are not exported in a way that is exposed to the browser.

## Solution
This PR changes the TS config module type to `ES6` for all our ESM configs and ensure module resolution is set to `bundler` which allows us to correctly find all our dependencies AND export the types.

Note: I also had to fix the linter, which does not understand module resolution `bundler`.

Built all the libraries as CJS and ESM and confirmed they build, lint, and display in the browser.


lukehb:
@mcottontensor This PR once landed will subsequently require releasing version bumps of common, signaling, library, and ui library in 5.5. Can I ask you to do that when landing?

mcottontensor:
Can do

Issue #21, opened 2024-02-29
Hide the QPIndicator when the connexion quality is good and the control div when in fullscreen


DenisTensorWorks:
Originally opened by [@TxMat](https://github.com/TxMat)

**Component your feature request relates to**
- Frontend UI library

**Problem your feature solves? Please describe.**
While the default UI being very clean already, the persistant presence of the controls button and Video QP Indicator even in fullscreen causes a lot of clutter on the game view and can be distracting for players.

**Describe the solution you'd like**
It'll be nice to have options to tweak to behaviors of these components by hiding them in full screen or only show the connexion indicator when the quality is bad to alert the player.

**Additional context**
I've tried implementing these changes myself by [tinkering](https://github.com/TxMat/PixelStreamingInfrastructure/tree/UE5.3) with the library
 *disclaimer : i am not a typescript dev and my changes are probably a bit wonky*
 
 ---
 
 Moved from https://github.com/EpicGames/PixelStreamingInfrastructure/issues/496

lukehb:
When we get round to implementing this we should add new frontend settings for each of these requests. (This is more modular than just hiding all the UI)

PR #405, opened 2025-01-16
New options to hide UI elements in certain situations

mcottontensor:
## Relevant components:
- [ ] Signalling server
- [ ] Common library
- [ ] Frontend library
- [x] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
#21
- We would like the option to hide the QP indicator when the connection is good.
- We would like to hide the controls when in fullscreen mode.

## Solution
This just adds a two new options to the application ui settings that allows the qp indicator to be hidden when the connection is good and the controls to be hidden when going fullscreen.
See.
`UIOptions.videoQPIndicatorConfig.hideWhenGood`
and
`UIOptions.hideControlsInFullscreen`

### DIFF
--- a/Frontend/ui-library/src/Application/Application.ts
+++ b/Frontend/ui-library/src/Application/Application.ts
@@ -46,6 +46,7 @@ import { FullScreenIconBase, FullScreenIconExternal } from '../UI/FullscreenIcon
  */
 export type VideoQPIndicatorConfig = {
     disableIndicator?: boolean;
+    /** Setting this to true will hide the indicator when the connection is "good" ie. green */
     hideWhenGood?: boolean;
 };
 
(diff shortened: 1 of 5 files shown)


Frontend/ui-library/src/Application/Application.ts
```diff
@@ -46,6 +46,7 @@ import { FullScreenIconBase, FullScreenIconExternal } from '../UI/FullscreenIcon
  */
 export type VideoQPIndicatorConfig = {
     disableIndicator?: boolean;
+    hideWhenGood?: boolean;
```
lukehb, line 50: Maybe a comment above this, saying something like:

When this flag is enabled we will hide the QP indicator when the connection is "good" (e.g. when the icon would normally be green).

lukehb (changes requested):
Minor suggestion about adding a comment.

mcottontensor:
Closes #21

Issue #34, opened 2024-03-19
Add Heartbeat to websocket connection between Frontend and SignallingServer


rotor-antoine:
**Component your feature request relates to**
- [x] Signalling server
- [x] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] SFU
- [ ] This repo

**Problem your feature solves? Please describe.**
The Frontend Library is not triggering disconnection event when I for instance turn off the wifi of my laptop (which is running the frontend).
Same issue on the Signalling Server. Those kind of "disconnection" wouldn't trigger an event that could be handle by the signalling server . 
I know that this repo is not supposed to be necessarily "production ready" but this issue could almost be considered a "bug" as the signalling server will never recover (ie it will never be availalble for a new user) when this type of disconnection with the frontend happen. 

**Describe the solution you'd like**
It seems like a hearbeat check (regular ping messages) both from frontend and from the signalling server, would be able to detect those disconnections:
https://github.com/websockets/ws?tab=readme-ov-file#how-to-detect-and-close-broken-connections

**Additional context**

lukehb:
This is a good idea.

I think this is a reasonable suggestion to include in our 5.5 release for the signaling server.

marospekarik:
This would also fix my problem with cloudfare terminating inactive ws connections after 90s. +1

lukehb:
Closed, landed in: https://github.com/EpicGamesExt/PixelStreamingInfrastructure/pull/409

PR #409, opened 2025-01-17
Adding a heartbeat to the signalling protocol.

mcottontensor:
## Relevant components:
- [X] Signalling server
- [X] Common library
- [X] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
#34
Sometimes a connection between the frontend and signalling server can be broken with no actual close event. With a heartbeat we would be able to timeout a connection and allow the user to retry.

## Solution
This just cleans up the ping/pong behaviour to implement it at the protocol level. If these messages don't come through in an expected amount of time the connection is closed and a special timeout event is fired.


Common/src/Protocol/SignallingProtocol.ts
```diff
@@ -24,27 +24,48 @@ import * as MessageHelpers from '../Messages/message_helpers';
...
+ *   timeout:
+ *      Emitted when a ping fails to receive a pong in time and the connection is
+ *      assumed to be dead. This is emitted after the connection is closed.
  */
 export class SignallingProtocol extends EventEmitter {
     static get SIGNALLING_VERSION(): string {
         return '1.2.1';
     }
 
+    static readonly PING_TIMEOUT: number = 30 * 1000;
```
lukehb: This feels like it should be configurable in:

1) frontend PS code
2) in SS config

The default seems good, but I can imagine people may have cases where they need it set to other values for some particular reason.

Common/src/Protocol/SignallingProtocol.ts
```diff
@@ -90,4 +112,84 @@ export class SignallingProtocol extends EventEmitter {
...
+
+        // mark the connection as temporarily dead until we get a response from the ping
+        this.alive = false;
+        this.sendMessage(MessageHelpers.createMessage(Messages.ping, { time: new Date().getTime() }));
+    }
+
+    private heartbeatWait(): void {
+        // non-initiator only.
+        // starts a timer to wait for heartbeat pings. we give it double the ping time to allow some lag
+        this.keepalive = setTimeout(this.heartbeatTimeout.bind(this), SignallingProtocol.PING_TIMEOUT * 2);
```
lukehb: As long as the user can set the value, I feel like it should just timeout to what the users sets, rather than `2*value`.

Additionally, I think if the user sets the timeout to zero or a negative value, then no timeout should be set/fired.

lukehb (changes requested):
Some minor changes requested, additionally if we the let the user set the value, there should be a slight docs update on the SS readme page when it mentions all the params.

Common/src/Protocol/KeepaliveMonitor.ts
```diff
@@ -0,0 +1,78 @@
...
+    private keepalive?: ReturnType<typeof setInterval>;
+    private alive: boolean = false;
+    private rtt: number = 0;
+
+    private onResponse: (pongMsg: Messages.pong) => void;
+
+    /**
+     * Called when a pong does not come back from a ping.
+     */
+    onTimeout?: () => void;
```
lukehb, line 23: Maybe this should be an EventEmitter type from Common, so we can have multiple listeners?

Common/src/Protocol/KeepaliveMonitor.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    private start(): void {
+        this.alive = true;
+        this.protocol.on('pong', this.onResponse);
+        this.keepalive = setInterval(this.sendHeartbeat.bind(this), this.timeout);
+    }
+
+    private stop(): void {
+        clearInterval(this.keepalive);
+        this.protocol.off('pong', this.onResponse);
```
lukehb, line 54:
```suggestion
        this.protocol.off('pong', ()=>{ this.onHeartbeatResponse });
```

Seems like `onResponse` is always `onHeartbeatResponse` , if so, let's just use `onHeartbeatResponse` directly and drop `onResponse`?

Common/src/Protocol/KeepaliveMonitor.ts
```diff
@@ -0,0 +1,78 @@
...
+        this.protocol = protocol;
+        this.timeout = timeout;
+        this.onResponse = this.onHeartbeatResponse.bind(this);
+        this.protocol.transport.on('close', this.stop.bind(this));
+        this.start();
+    }
+
+    private start(): void {
+        this.alive = true;
+        this.protocol.on('pong', this.onResponse);
```
lukehb, line 48: I think use of arrow syntax over bind/lack of bind is usually the safer choice.

```suggestion
        this.protocol.on('pong', ()=> { this.onHeatbeatResponse() });
```

Common/src/Protocol/KeepaliveMonitor.ts
```diff
@@ -0,0 +1,78 @@
...
+    private stop(): void {
+        clearInterval(this.keepalive);
+        this.protocol.off('pong', this.onResponse);
+    }
+
+    private sendHeartbeat(): void {
+        // if we never got a response from the last heartbeat, assume the connection is dead and timeout
+        if (this.alive === false) {
+            this.onTimeout?.();
+            // this.protocol.disconnect();
```
lukehb: I think this is the safest choice.

lukehb (changes requested):
Few minor suggestions about event handling and binding

Common/src/Protocol/KeepaliveMonitor.ts
mcottontensor: I did it this way so we can `on` and `off` the callback which I can't do if i use bind or an unnamed lambda.

Common/src/Protocol/KeepaliveMonitor.ts
mcottontensor: As mentioned above I create a named lambda in onResponse so that I can reference it when using `on` and `off`

Common/src/Protocol/KeepaliveMonitor.ts
mcottontensor: I was debating this but, in my mind at least, the normal use for this is that whoever creates this is only interested in one thing and the need for multiple listeners is minimal. If you do need to notify more things it makes sense to me that the creator of this object would have the responsibility of notifying things.

lukehb (approved):

PR #410, opened 2025-01-17
Fix: Chrome + iPad not creating rootElement early enough for custom frontends

lukehb:
## Relevant components:
- [x] Frontend UI library

## Problem statement:
The `_rootElement` in the Pixel Streaming UI library was not created explicitly, but rather created lazily by a getter called `rootElement`. There were some scenarios where you could trigger internal code paths that used the `_rootElement` before any call had been made via the getter to initialize it. This would result in a JS error for those calling functions.

## Solution
This PR explicitly initializes the `_rootElement` in the constructor of the Pixel Streaming `Application` - this way the calling order is no longer important and it is always safe to access the `_rootElement`.

I have tested this locally using my Chrome browser, but did not have an iPad nearby to confirm. I will get someone to test this downstream once it is landed to confirm the fix is valid.

### DIFF
--- a/Frontend/ui-library/src/Application/Application.ts
+++ b/Frontend/ui-library/src/Application/Application.ts
@@ -112,6 +112,13 @@ export class Application {
         this._options = options;
 
         this.stream = options.stream;
+
+        // Explicitly create ui features now so creation time is known
+        this._uiFeatureElement = this.createUIFeaturesElement();
+
+        // Explicitly create root element now so creation time is known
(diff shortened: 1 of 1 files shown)


mcottontensor (approved):

mcottontensor:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|UE5.3|[](https://github.com/EpicGamesExt/PixelStreamingInfrastructure/pull/412)|
|✅|UE5.4|[](https://github.com/EpicGamesExt/PixelStreamingInfrastructure/pull/413)|

### Questions ?

Issue #332, opened 2024-11-27
Implement a way to automatically rebuild components whenever typescript code is changed


DenisTensorWorks:
**Component your feature request relates to**
- [x] Build scripts

**Problem your feature solves? Please describe.**
Currently whenever a user makes changes to the typescript code, the user needs to be rebuild that code and potentially any dependent libraries, this can be quite high friction to dev iteration given our monorepo has quite a few packages now.

It would be nice to have a smoother developer iteration cycles, for example, if it could auto detect changes to any of the dependent libraries and rebuild all relevant components, something akin to:

`cd ./Frontend/implementations/typescript`
`npm run watch`

mcottontensor:
Just dumping down this
https://lerna.js.org/
I haven't actually looked into it yet. Just saw mention of it and want to look into it when I get back to this task.

Same deal with https://nx.dev/

lukehb:
Also dumping some idea here:

The high level idea is:

- Each package has its own npm script to start the webpack dev server (that already does watching for us)

- By creating an uber script that starts all the dev servers we should end up in a situation where you can make a change in common and it will rebuild common, all the things that depend on common, and all the things that depend on those things


It would be pretty nice in theory, basically go to repo root and run something like `npm run dev`

Then can just make changes wherever and things will get rebuilt.

lukehb:
Another thought I had that is about speeding up dev iteration:

We should ensure all our packages have a `compile-dev` and `build-dev` option that only does one of the compilation options (`cjs` or `esm` - not both). While compiling both is good for when we go to publish, for dev, we really only want one - doing both at every package in the repo makes the compilation take twice as long.

tldr; cut down iteration time by only doing half the compiling when we know we are devving.

mcottontensor:
The majority of this is in now with the `develop` script for SS. There are some improvements that can and will probably need to be done but they can be new tasks.

PR #420, opened 2025-01-22
Adding a 'watch' mode to most projects

mcottontensor:
## Relevant components:
- [x] Signalling server
- [x] Common library
- [x] Frontend library
- [x] Frontend UI library
- [ ] Matchmaker
- [x] Platform scripts
- [x] SFU

## Problem statement:
#332
The build process is finicky and error prone with knowing what to build and when. Some build scripts are even broken in some situations. We want to automate or semi-automate the build process so that development is more streamlined.

## Solution
After much fussing about with mono repo tools and webpack etc. I found that using a series of nodemon scripts along with tsc watch mode and webpack was the simplest solution.
There is now a single script called `develop` in the signalling server project that will go though all the components and kick off a watch process that will not only watch for source changes but also dependency changes and auto build and when needed restart the signalling server.
This should allow you to start the develop script and then make changes to whatever component you need, including frontend changes, and have them working and served immediately.

## Documentation
Updated SS launch docs mentioning `npm run develop`

See above.


Common/package.json
```diff
@@ -1,49 +1,55 @@
...
+        "build:docs:proto": "protoc --doc_out=docs --doc_opt=markdown,messages.md --proto_path protobuf protobuf/signalling_messages.proto",
+        "build:docs:ts": "typedoc --tsconfig tsconfig.base.json",
+        "build:docs": "npm run build:docs:proto && npm run build:dos:ts",
+        "build:proto": "protoc --experimental_allow_proto3_optional --ts_out src/Messages --proto_path protobuf protobuf/signalling_messages.proto",
+        "build:cjs": "tsc --project tsconfig.cjs.json",
+        "build:esm": "tsc --project tsconfig.esm.json",
+        "build": "npm run build:proto && npm run build:cjs && npm run build:esm",
+        "rebuild": "npm run clean && npm run build",
+        "watch:proto": "nodemon -V -d 3 --watch protobuf/signalling_messages.proto --exec 'npm run build:proto'",
+        "watch": "concurrently -k \"npm run watch:proto\" \"tsc --watch --preserveWatchOutput --project tsconfig.esm.json\"",
```
lukehb, line 20: Where does the `concurrently` command come from?

mcottontensor: It's an npm package.

Common/tsconfig.esm.json
```diff
@@ -1,8 +1,8 @@
 {
     "extends": "./tsconfig.base.json",
     "compilerOptions": {
-        "outDir": "./build/esm",
-        "module": "ES6",
-        "moduleResolution": "bundler"
+        "outDir": "./dist/esm",
+        "module": "nodenext",
```
lukehb: Ah... this is a complicated one to explain, but I am _pretty_ sure it has to build as `"module: ES6"` to produce a library that can be webpacked into an ES6 bundle.

For instance if we were to change the webpack in `Frontend\implementations\typescript\webpack.common.js` to produce a ES6 module instead of commonjs and you tried to bundle the built `Frontend/implementations/typescript/package.json` and run `npm run build-esm` on it

then try to use that module like:


```
<script type=module src="./player.esm.js">
import * as PS from './player.esm.js';
console.log(PS);
</script>
```

Here is the nuance, `module: nodenext` WILL bundle, but it doesn't produce JS that results in a webpacked ES6 bundle that has exported functions.

That is, that log above would not include the exported functions of the common or other libs unless they specifically built in TS as `ES6`. This is a limitation of webpack, but a nuanced one that means `module: ES6` was load bearing. I should probably have written a test to capture this requirement.

I assume there was a reason to change it to `nodenext` though?

Frontend/ui-library/package.json
```diff
@@ -1,44 +1,47 @@
...
+    "module": "dist/esm/pixelstreamingfrontend-ui.esm.js",
+    "types": "dist/types/pixelstreamingfrontend-ui.d.ts",
+    "sideEffects": false,
+    "scripts": {
+        "clean": "rimraf ./dist",
+        "build:cjs": "tsc --project tsconfig.cjs.json",
+        "build:esm": "tsc --project tsconfig.esm.json",
+        "build": "npm run build:cjs && npm run build:esm",
+        "rebuild": "npm run clean && npm run build",
+        "watch": "nodemon -V -d 3 --watch src --watch ../library/dist -e 'ts,js,mjs,cjs,json' --exec 'npm run build:cjs'",
```
lukehb, line 15: Why build cjs for this one but esm for base library?

SignallingWebServer/package.json
```diff
@@ -2,14 +2,16 @@
...
         "build": "tsc",
-        "build-dev": "npm run build",
-        "build-all": "rimraf ./build && cd ../Signalling && npm run build-all && cd ../SignallingWebServer && npm run build-dev",
-        "start": "node ./build/index.js --serve --console_messages verbose --log_config --https_redirect",
+        "rebuild": "npm run clean && npm run build",
+        "start": "node ./dist/index.js --serve --console_messages verbose --log_config --https_redirect",
         "lint": "eslint ./src --ext .js,.jsx,.ts,.tsx",
-        "test": "echo \"Error: no test specified\" && exit 1"
+        "test": "echo \"Error: no test specified\" && exit 1",
+        "watch": "nodemon -V -d 3 --watch src --watch ../Signalling/dist -e 'ts,js,mjs,cjs,json' --exec 'npm run build && node ./dist/index.js --serve --console_messages verbose --log_config --https_redirect --player_port 1025'",
```
lukehb: I guess player port 1025 is beause of elevated perms required on Linux, this means user should visit `http://localhost:1025` when using `watch`? Let's add this to our docs then.

lukehb (changes requested):
Some comments about what I believe to be load bearing configuration of how ESM modules were being built.

Additionally, we should land docs changes for how to use `watch` in this PR too.

Besides that, PR and approach looks good!

Common/tsconfig.esm.json
mcottontensor: It can probably be changed back. It changed somewhere in the 400 attempts to get various build systems working.

Frontend/ui-library/package.json
mcottontensor: Honestly I have no idea. It was the one that worked in triggering changes. Maybe related to the nodenext change.

SignallingWebServer/package.json
mcottontensor: Yes. Dealing with the elevated perms was just not worth it since using sudo changed the environment and npm had to be available in that environment even if I tried to use -E that is supposed to keep the environment I was still getting npm not found issues.

Common/tsconfig.esm.json
lukehb: Lmk if changing it back causes issues, I am happy to help out on this as it is nuanced and also fresh for me due to working on similar issues elsewhere.

Frontend/ui-library/package.json
lukehb: My intuition is they should either all be cjs or all esm (while a mixture will work, we may as well be consistent as it will help webpack tree shaking etc).

Common/package.json
lukehb: I think I have code blindness, I cannot see it in the dev deps or deps of this package?

Frontend/ui-library/package.json
mcottontensor: So looking into it further. It seems that the typescript implementation project, the webpack.dev config file doesnt overwrite the tsconfig.cjs config reference. So it seems its using the cjs version of the ui library.

lukehb: I am happy for them to all be commonjs in that case, it is probably closest to what we have shipped in the past (e.g. safest in terms of any potential differences creeping in).

Note, we use the implementation/typescript directly in things like VCam so I would like to keep it as close to what it was before as possible.

From memory the implementation/typescript consumes them as commonjs, then itself produces a `umd` bundle for maximum compatibility - which should also be preserved.

mcottontensor: In playing around with it, I cant for the life of me figure out why webpack is using the cjs version of ui-library but everything else seems to be esm. I poked around with various configs but nothing made it work.

lukehb: In the end, as long as it is matching whatever it was before this PR it should be safe.

Frontend/implementations/typescript/package.json
```diff
@@ -1,32 +1,32 @@
...
-  "main": "./src/player.ts",
-  "scripts": {
-    "build": "npx webpack --config webpack.prod.js",
-    "build-dev": "npx webpack --config webpack.dev.js",
-    "build-esm": "npx webpack --config webpack.esmodule.js",
-    "watch": "npx webpack --watch",
-    "serve": "webpack serve --config webpack.dev.js",
-    "serve-prod": "webpack serve --config webpack.prod.js",
-    "build-all": "cd ../../ui-library && npm run build-all && cd ../implementations/typescript && npm install && npm run build",
-    "build-dev-all": "cd ../../ui-library && npm run build-all && cd ../implementations/typescript && npm install && npm run build-dev"
```
lukehb, line 14: Idea with removing `build-all` etc is that you can do `npm run build` in the repo root or `watch` now?

mcottontensor: It was just another point of failure mostly. Yes you can do build on the workspace, but most of the time I imagine people will use the platform scripts which already build whats needed.
I'm debating on whether the `develop` script should be moved into the root too so that all npm scripts in a project only relate to the project only.

Common/package.json
lukehb: So the dep is declared in SS, but it works here too. I'm both amazed and confused.

lukehb: I feel like dep should be declared here too for future proofing, e.g. imagine if it is removed from SS but still needed here.

mcottontensor: Yeah this has been an issue with going to the workspaces way of doing things. I regularly use things that exist because they're a dep in another package and theres no issue, until packages are cleared or something and suddenly it's an issue.
Finding these issues is difficult.

lukehb: Agreed

lukehb: I think let's add this to the package.json in common and I will approve

mcottontensor: Done

lukehb (approved):

mcottontensor:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|UE5.5|[](https://github.com/EpicGamesExt/PixelStreamingInfrastructure/pull/433)|

### Questions ?

PR #425, opened 2025-01-30
Add: Latency measurements such as sender to receiver latency measurements using RTP header extensions abs-capture-time and video-timing

lukehb:
## Relevant components:
- [x] Common library
- [x] Frontend library
- [x] Frontend UI library

## Problem statement:
Adds optional [abs-capture-time](https://www.ietf.org/archive/id/draft-alvestrand-avtcore-abs-capture-time-00.html) RTP header extension to the negotiated SDP (only works with Chromium+audio as of now). Additionally use the [video-timing](https://webrtc.github.io/webrtc-org/experiments/rtp-hdrext/video-timing/) header extension (only works in Chromium+video as of now) to get video statistics from both sender and receiver time including: encoding delay, packetizer delay, pacer delay.

## Solution

1. Adds a new config flag to enable `abs-capture-time` to RTP header extension
2. Munges the extension into the SDP
3. Adds appropriate tests to use this header extension to measure latency using [captureTimestamp](https://w3c.github.io/webrtc-extensions/explainer.html).
4. Adds some more latency stats to the frontend stats panel to display the moment to moment latency using information from both abs-capture-time (if possible) and video-timing.

image
 
**Note:** these only show up in supported browsers - that is, much of this does nothing on Firefox because many of these code paths are completely skipped on Firefox.

## Documentation
N/A - The flag is enabled by default and not exposed to the UI. It should be a transparent change to user and should not need configuration.

All unit tests, playwright tests, and CI have been updated accordingly (this was tricky as testing revealed Firefox refuses to connect if this flag there are too many header extensions so this is carefully skipped on Firefox).


lukehb:
Based on this: https://github.com/EpicGamesExt/PixelStreamingInfrastructure/actions/runs/13026995654

@mcottontensor  Landing this PR would require new `common` and `lib-pixelstreamingfrontend` versions to be cut.

.github/workflows/healthcheck-libraries.yml
```diff
@@ -57,4 +57,8 @@ jobs:
 
     - name: Clean build of frontend implementation
       working-directory: Frontend/implementations/typescript
-      run: npm install && npm run build
\ No newline at end of file
+      run: npm install && npm run build
+
+    - name: Clean build of frontend implementation as ES6
```
lukehb, line 62: Added this step to the CI, as this was actually broken and not being caught.

.github/workflows/healthcheck-streaming.yml
```diff
@@ -60,11 +60,11 @@ jobs:
       uses: robinraju/release-downloader@v1
       with:
         repository: 'EpicGamesExt/PixelStreamingInfrastructure'
-        tag: 'minimal-streamer'
-        fileName: 'Minimal-PixelStreamer-5.5.7z'
+        tag: 'minimal-streamer-5.5'
+        fileName: 'Minimal-PixelStreamer-5.5-Win64-Development.7z'
```
lukehb, line 64: Upgraded to using our latest minimal streamer release, as the other one was a bit out of date now.

Extras/FrontendTests/playwright.config.ts
```diff
@@ -14,8 +14,8 @@ export default defineConfig({
   forbidOnly: !!process.env.CI,
   /* Retry on CI only */
   retries: process.env.CI ? 2 : 0,
-  /* Opt out of parallel tests on CI. */
-  workers: process.env.CI ? 1 : 5,
+  /* Opt out of parallel tests in general as multiple streamers mean they can connect to the wrong test */
+  workers: 1,
```
lukehb, line 18: This was making tests flaky in CI

Extras/FrontendTests/tests/basic_stream.spec.ts
```diff
@@ -10,20 +11,22 @@ test('Test default stream.', {
...
+    await helpers.startAndWaitForVideo(page);
+
+    let frameCount: number = await page.evaluate(()=> {
+        return new Promise<number>((resolve) => {
+            window.pixelStreaming.addEventListener("statsReceived", (e: StatsReceivedEvent) => {
+                if(e.data.aggregatedStats && e.data.aggregatedStats.inboundVideoStats && e.data.aggregatedStats.inboundVideoStats.framesReceived) {
+                    resolve(e.data.aggregatedStats.inboundVideoStats.framesReceived);
+                }
+            });
+        });
```
lukehb, line 25: I changed this to be callback based instead of asking for stats straight away to fix test flakiness where it would ask for stats rarely before there was any

Extras/FrontendTests/tests/fixtures.ts
```diff
@@ -3,25 +3,39 @@ import { test as base, Page } from '@playwright/test';
...
+            // to be called when the endpoint_id_confirmed is actually called.
+            streamerPage.evaluate(()=> {
                 window.streamer.on('endpoint_id_confirmed', () => {
-                    resolve(window.streamer.id);
+                    window.resolveFromIdPromise(window.streamer.id);
                 });
-            })
+            });
         });
+
```
lukehb, line 32: Without this reordering I would very rarely get a test hang

Extras/FrontendTests/tests/keyboard.spec.ts
```diff
@@ -12,8 +12,8 @@ test('Test keyboard events', {
 }, async ({ page, streamerPage, streamerId }) => {
 
     await page.goto(`/?StreamerId=${streamerId}&MatchViewportRes=true`);
-    await page.getByText('Click to start').click();
-    await helpers.waitForVideo(page);
+
+    await helpers.startAndWaitForVideo(page);
```
lukehb, line 16: Previously where it would click and then call wait for video, sometimes it would execute the whole thing in the call to click and `playStream` event that the wait for video waits on would already be fired. This change ensures the listening to the event is done before it does the starting of the stream.

Extras/JSStreamer/Dockerfile
```diff
@@ -1,9 +1,22 @@
 FROM node:20-bookworm
 
+## Note: This dockerfile is expected to be called from the root of this repo
+## Maybe something like: docker build -t epicgames/jsstreamer:latest -f ./Extras/JSStreamer/Dockerfile .
+
 WORKDIR /streamer
-COPY /Extras/JSStreamer .
+COPY /Common ./Common
+COPY /Extras/JSStreamer ./Extras/JSStreamer
+COPY ./package.json ./package.json
```
lukehb, line 9: Ensure streamer has the deps from the local repo as opposed to remote/published deps that may not have the new changes we want to test.

.github/workflows/healthcheck-streaming.yml
```diff
@@ -121,9 +121,16 @@ jobs:
     - name: Wait for signalling to come up
       run: curl --retry 10 --retry-delay 20 --retry-connrefused http://localhost:999/api/status
 
+    - name: Wait for front end to come up
+      run: curl --retry 10 --retry-delay 20 --retry-connrefused http://localhost:999/
+
```
mcottontensor, line 124: I don't know if this is needed. Since the api comes from the frontend it should be testing both, no?

.github/workflows/healthcheck-streaming.yml
```diff
@@ -121,9 +121,16 @@ jobs:
...
+    - name: Wait for front end to come up
+      run: curl --retry 10 --retry-delay 20 --retry-connrefused http://localhost:999/
+
     - name: Wait for streamer to come up
       run: curl --retry 10 --retry-delay 20 --retry-connrefused http://localhost:999/api/streamers/DefaultStreamer
 
+    - name: Output streamer logs
+      working-directory: Streamer
+      run: ls ".\Minimal\" && Test-Path ".\Minimal\Saved\Logs\Minimal.log" && cat ".\Minimal\Saved\Logs\Minimal.log"
+
```
mcottontensor, line 131: Is this needed?

Extras/FrontendTests/playwright.config.ts
mcottontensor: Should the change be defining CI as an env variable?

Extras/FrontendTests/tests/helpers.ts
```diff
@@ -31,12 +31,21 @@ export function delay(time: number) {
   });
 }
 
-export async function waitForVideo(page: Page) {
+export async function startStreaming(page: Page) {
+    await page.evaluate(()=> {
+        window.pixelStreaming.connect();
+    });
+}
```
mcottontensor: Is this not used?

.github/workflows/healthcheck-streaming.yml
lukehb: It is not _needed_ but I found it useful to verify UE side bugs in CI. I can remove it if you prefer?

.github/workflows/healthcheck-streaming.yml
```diff
@@ -121,9 +121,16 @@ jobs:
     - name: Wait for signalling to come up
       run: curl --retry 10 --retry-delay 20 --retry-connrefused http://localhost:999/api/status
 
+    - name: Wait for front end to come up
```
lukehb:
```suggestion
```

.github/workflows/healthcheck-streaming.yml
```diff
@@ -121,9 +121,16 @@ jobs:
     - name: Wait for signalling to come up
       run: curl --retry 10 --retry-delay 20 --retry-connrefused http://localhost:999/api/status
 
+    - name: Wait for front end to come up
+      run: curl --retry 10 --retry-delay 20 --retry-connrefused http://localhost:999/
```
lukehb:
```suggestion
```

Extras/FrontendTests/playwright.config.ts
lukehb: I don't think so. The way the tests work they rely on an external SS and JSStreamer to be run. This means if two tests ever run in parallel they will both connect to these external services. This was messing up the tests in some cases. Without making it so each tests starts its own SS and JSStreamer the easiest fix was simply to make the tests completely serial (by limiting the workers to 1 in all cases).

Extras/FrontendTests/tests/helpers.ts
lukehb: Good point, I think all the tests now call `startAndWaitForVideo` ?

lukehb:
```suggestion
```

Frontend/implementations/typescript/tsconfig.esm.json
```diff
@@ -2,7 +2,7 @@
     "extends": "./tsconfig.base.json",
     "compilerOptions": {
         "outDir": "./dist/esm",
-        "module": "es6",
-        "moduleResolution": "bundler"
+        "module": "ES6",
+        "moduleResolution": "nodenext"
```
lukehb:
```suggestion
        "module": "es6",
        "moduleResolution": "bundler"
```

lukehb:
~~Note to self: I need to merge changes from master into here to fix CI and additionally I need to test this change on more browser and devices (iOS / Android / Safari / FF).~~ Done!

.github/workflows/healthcheck-streaming.yml
lukehb: I removed this in the latest commit

lukehb:
Merged master in now.

~~Still need to test on devices/browser~~ Done! All good.

 ~~Also want to make which capture to send method is used clearer. This is because both `video-timing` and `abs-capture-time` both _can_ calculate it, but it is unclear to me if these are the same/similar due to Chrome not yet supporting remote-outbound-rtp stats for video. Because Chrome does not support these stats, I can't calculate capture to send for video using abs-capture-time. Therefore, the current commit falls back to using video timing to calculate capture to send time, but we should make this clearer this approach is used.~~ Done: https://github.com/EpicGamesExt/PixelStreamingInfrastructure/pull/425/commits/ee5d7bd61f6a7ac48d2bdc828311aa9275d08922

Extras/FrontendTests/tests/helpers.ts
lukehb: Updated, so the unused one was removed.

lukehb:
@mcottontensor When we go to cut a new version from this, there is some API breakages due to some changing of stats casing. So I think we have bump the major version number.

mcottontensor (approved):

mcottontensor:
Do we want to do a full release? Or just common and the frontend libs?

mcottontensor:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|UE5.5|[](https://github.com/EpicGamesExt/PixelStreamingInfrastructure/pull/449)|

### Questions ?

Frontend/library/src/Config/Config.ts
```diff
@@ -564,6 +565,19 @@ export class Config {
...
 
+        this.flags.set(
+            Flags.EnableCaptureTimeExt,
+            new SettingFlag(
+                Flags.EnableCaptureTimeExt,
+                'Enable abs-capture-time',
+                'Enables the abs-capture-time RTP header extension',
+                settings && Object.prototype.hasOwnProperty.call(settings, Flags.EnableCaptureTimeExt)
+                    ? settings[Flags.EnableCaptureTimeExt]
+                    : true,
```
rlamarche, line 576: Hi, is it voluntary to enable this by default ? It breaks compatiblity with not up to date Chrome/Chromium browser.

lukehb: @rlamarche Thanks for reporting, I have put together a PR: https://github.com/EpicGamesExt/PixelStreamingInfrastructure/pull/516

PR #429, opened 2025-02-05
Fixing build configs. Tweaking Protocol message usage. Logging. Docs.

mcottontensor:
## Relevant components:
- [x] Signalling library
- [x] Common library
- [x] Frontend library
- [x] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
There were several issues with the build setup relating to ESM and CJS builds where in some configurations libraries would fail to resolve.
The usage of Protocol Messages was a little confusing.
Documentation on the signalling protocol was split over common and signalling.

## Solution
I have updated the build configuration so it should be possible to now build esm or cjs in isolatiion and have it all work.
I moved the usage of protocol messages out of transport and into signalling protocol.
I have moved the signalling documentation into common.
I have updated the logger usage in common so it can be overridden since the Logger in common and Logger in signalling handle things differently. It's not the best solution but its a solution for now.
I have rebuilt the docs for both common and signalling so it should be up to date.

## Documentation
Built.


Frontend/ui-library/tsconfig.base.json
```diff
@@ -10,7 +10,9 @@
         "noImplicitOverride": true,
         "noFallthroughCasesInSwitch": true,
         "noImplicitReturns": true,
-        "noPropertyAccessFromIndexSignature": true
+        "noPropertyAccessFromIndexSignature": true,
```
lukehb, line 13: Interesting, was there issues without this?

package.json
```diff
@@ -18,6 +18,8 @@
     "private": true,
     "scripts": {
         "build": "npm run build --ws",
+        "build:all:cjs": "cd Common && npm run build:cjs && cd ../Signalling && npm run build:cjs && cd ../SignallingWebServer && npm run build && cd ../Frontend/library && npm run build:cjs && cd ../ui-library && npm run build:cjs && cd ../implementations/typescript && npm run build:dev",
```
lukehb, line 21: Another option could be `npm run build:cjs --ws` as long as every package has that option (even if is a no-op in `Extras/`). I think this is probably better for specificity though and order of operations.

lukehb (approved):
Approved, left a few comments and questions. But no changes required, looks good to merge,

Frontend/ui-library/tsconfig.base.json
mcottontensor: I didn't change it, just added the comma.

package.json
mcottontensor: Well the order is important. I assume the --ws just follows the order in the workspace entries but thats kind of loose. We also have some other packages in the repo that arent required for development.

lukehb: Agreed

mcottontensor:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|UE5.5|[](https://github.com/EpicGamesExt/PixelStreamingInfrastructure/pull/438)|

### Questions ?

PR #476, opened 2025-02-23
Adding the branch name to the publish tags

### LINKED ISSUE #470: Chore: Github tags on packages do not include UE suffix
**Component your issue relates to**
- [x] This repo

**Problem your feature solves? Please describe.**
GH tags for our packages don't include the UE version in the name. This is a bit of a problem when we make the UE 5.6 branch for example because we reset the version numbers on all our packages in the past and have given them a new name with the UE5.X suffix.

https://github.com/EpicGamesExt/PixelStreamingInfrastructure/tags

Without proper naming there will be clash once UE 5.6 packages start being created.

**Describe the solution you'd like**
Add the UE suffix to these package tags (manually fix the already created ones) and fix up the action that creates them to have the UE suffix going forward.

mcottontensor:
#470

This should properly add the UE version to the publish tags so they dont conflict on new UE releases.

### DIFF
--- a/.github/workflows/publish-common-library-to-npm.yml
+++ b/.github/workflows/publish-common-library-to-npm.yml
@@ -10,7 +10,7 @@ env:
 
 jobs:
   build:
-    if: github.repository == 'EpicGamesExt/PixelStreamingInfrastructure'
+    # if: github.repository == 'EpicGamesExt/PixelStreamingInfrastructure'
     runs-on: ubuntu-latest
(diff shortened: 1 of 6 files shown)


lukehb (approved):

lukehb:
Will need backporting to 5.5.

Also, all our existing tags for releases of 5.5 libs will need manually renaming to include the UE5.5 prefix as you have done here.

PR #509, opened 2025-02-25
Fix: Git bash failing to run platform scripts on Windows due to tar conflict

### LINKED ISSUE #473: [BUG] - Setup error with TAR unzip .bat
**Problem component**
Setup.Bat

**Description**
When running Setup.bat the system attempts to unzip the node.zip file but TAR returns the following error:

tar (child): Hmm, this doesn't look like a tar archive

This was resolved by modifying the common.bat file to add the following code to make setup.bat unzip the file using powershell instead:

 powershell -Command "Expand-Archive -Path 'node.zip' -DestinationPath '.' -Force"  

**Steps to Reproduce:**
1. Go to \SignallingWebServer\platform_scripts\cmd
2. Click on Setup.bat

**Expected behavior**
1. Node.zip unzips and continues with setup

**Screenshots**

Image

 - OS: [e.g. Windows 10 Pro]
 - Browser + version [e.g. chrome 100]

lukehb:
## Relevant components:
- [x] Platform scripts

## Problem statement:
Some users want to use Git bash on Windows to run the `platform_scripts/cmd`.
This currently fails as Git bash ships its own (gnu) `tar` which does not support `.zip` files.
Our intent is to use (bsd)`tar` that ships with Windows.

Closes https://github.com/EpicGamesExt/PixelStreamingInfrastructure/issues/473

## Solution
The solution is to explicitly call (bsd) tar which resides in `System32`

## Documentation
N/A

Tested with Git bash and cmd.exe both now work.

### DIFF
--- a/SignallingWebServer/platform_scripts/cmd/common.bat
+++ b/SignallingWebServer/platform_scripts/cmd/common.bat
@@ -122,7 +122,7 @@ if exist node\ (
   rem Download nodejs and follow redirects.
   curl -L -o ./node.zip "https://nodejs.org/dist/%NODE_VERSION%/%NODE_NAME%.zip"
 
-  tar -xf node.zip
+  %SystemRoot%\System32\tar.exe -xf node.zip
   ren "%NODE_NAME%\" "node"
   del node.zip
 )
@@ -206,7 +206,7 @@ if exist coturn\ (
   curl -L -o ./turnserver.zip "https://github.com/EpicGamesExt/PixelStreamingInfrastructure/releases/download/v4.5.2-coturn-windows/turnserver.zip"
 
   @Rem Unarchive the .zip to a directory called "turnserver"
-  mkdir coturn & tar -xf turnserver.zip -C coturn
+  mkdir coturn & %SystemRoot%\System32\tar.exe -xf turnserver.zip -C coturn
 
   @Rem Delete the downloaded turnserver.zip
   del turnserver.zip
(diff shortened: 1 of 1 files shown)


SignallingWebServer/platform_scripts/cmd/common.bat
```diff
@@ -206,7 +206,7 @@ if exist coturn\ (
   curl -L -o ./turnserver.zip "https://github.com/EpicGamesExt/PixelStreamingInfrastructure/releases/download/v4.5.2-coturn-windows/turnserver.zip"
 
   @Rem Unarchive the .zip to a directory called "turnserver"
-  mkdir coturn & tar -xf turnserver.zip -C coturn
+  mkdir coturn & %SystemRoot%\System32\tar.exe -xf turnserver.zip -C coturn
```
mcottontensor: Would prefer if the path was a variable. ie
```
set TAR="%SystemRoot%\System32\tar.exe"
```
up near where NPM is defined. Then just use `%TAR%`

mcottontensor (changes requested):

SignallingWebServer/platform_scripts/cmd/common.bat
lukehb: On it

lukehb: Done: https://github.com/EpicGamesExt/PixelStreamingInfrastructure/pull/509/commits/bf33b3c4d44fd81bd61bb6ee5e12aa29ec91287f

mcottontensor:
Will probably need separate changes for older branches than 5.5 since the scripts changed significantly.

lukehb:
> Will probably need separate changes for older branches than 5.5 since the scripts changed significantly.

It is okay, we'll leave those older branches as they are then

PR #518, opened 2025-03-03
Fixing the backspace code thats sent to UE

mcottontensor:
## Relevant components:
- [ ] Signalling server
- [ ] Common library
- [x] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
Backspace was not working for widget input.

## Solution
There's a fair amount of deprecated code in the input system that is what I think lead to this issue, but we have a special case for backspace so it gets sent to UE, only we weren't filling in the correct properties on the KeyboardEvent before sending it and so the event was never handled. I've now added the correct properties.

I've also done a small amount of cleanup and added some comments explaining a few things.

### DIFF
--- a/Frontend/library/src/Inputs/KeyboardController.ts
+++ b/Frontend/library/src/Inputs/KeyboardController.ts
@@ -19,7 +19,7 @@ export class KeyboardController implements IInputController {
     onKeyDownListener: (event: KeyboardEvent) => void;
     onKeyUpListener: (event: KeyboardEvent) => void;
     onKeyPressListener: (event: KeyboardEvent) => void;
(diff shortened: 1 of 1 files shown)


Belchy06 (approved):
LGTM!

lukehb:
@mcottontensor Can I request a new version of the 5.5 frontend library be cut to include this change also, so downstream users can update to the newest frontend lib?

PsiCondition:
Thanks a lot guys ! 

We will be waitting for the new version.

mcottontensor:
Pushed a new frontend library version today. 1.0.2

Issue #404, opened 2025-01-15
[Feature Request] Specify protocol when creating Websocket connection


fabian-rehm:
**Component your feature request relates to**
- [ ] Signalling server
- [X] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] SFU
- [ ] This repo

**Problem your feature solves? Please describe.**
The current implementation of the WebSocket connection in the Pixel Streaming library lacks the flexibility to specify a protocol when establishing a connection to the signaling server. This limitation can hinder compatibility with certain signaling servers that require subprotocols for proper communication.

**Describe the solution you'd like**
When creating a websocket with the library 'ws' you can first specify the URL and optional specify a protocol.

https://github.com/EpicGamesExt/PixelStreamingInfrastructure/blob/master/Common/src/Transport/WebSocketTransportNJS.ts#L46
https://github.com/EpicGamesExt/PixelStreamingInfrastructure/blob/UE5.4/Frontend/library/src/WebSockets/WebSocketController.ts#L40

At those 2 positions it would be possible to specify the protocol in the exact way we would need it.
this.webSocket = new WebSocket(connectionURL, **_protocol_**);

Best possible solution would be to have a settings parameter to specify this protocol when creating the PixelStream object.

lukehb:
For the implementor, please refer to: https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/WebSocket#protocols

And for examples of protocols that are known for use with websocket: https://www.iana.org/assignments/websocket/websocket.xml#subprotocol-name

mcottontensor:
Has been merged into master and UE5.5

PR #521, opened 2025-03-03
Websocket subprotocol support

mcottontensor:
## Relevant components:
- [ ] Signalling server
- [x] Common library
- [x] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
#404
We would like to be able to provide websocket protocols to the websocket creation.

## Solution
Added a configuration parameter that allows the user to specify the protocols to the websocket creation.

## Documentation
Updated the typedocs

### DIFF
--- a/Common/src/Transport/WebSocketTransportNJS.ts
+++ b/Common/src/Transport/WebSocketTransportNJS.ts
@@ -13,13 +13,33 @@ import WebSocket from 'ws';
 export class WebSocketTransportNJS extends EventEmitter implements ITransport {
     WS_OPEN_STATE = 1;
     webSocket?: WebSocket;
+    protocols?: string | string[];
 
-    constructor(existingSocket?: WebSocket) {
+    /**
+     * Constructs a new instance of a WebSocketTransport for node contexts.
+     * @param existingSocket - An existing WebSocket to use.
+     */
+    constructor(existingSocket: WebSocket);
+
+    /**
+     * Constructs a new instance of a WebSocketTransport for node contexts.
+     * @param protocols - A protocol or list of protocols to pass to the new websocket.
+     */
+    constructor(protocols: string | string[]);
+
+    /**
+     * Constructs a new instance of a WebSocketTransport for node contexts.
+     */
+    constructor();
+
+    constructor(argument?: WebSocket | string | string[]) {
         super();
-        if (existingSocket) {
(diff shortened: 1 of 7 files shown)


Common/src/Transport/WebSocketTransportNJS.ts
```diff
@@ -13,13 +13,33 @@ import WebSocket from 'ws';
...
     WS_OPEN_STATE = 1;
     webSocket?: WebSocket;
+    protocols?: string | string[];
 
-    constructor(existingSocket?: WebSocket) {
+    /**
+     * Constructs a new instance of a WebSocketTransport for node contexts.
+     * @param existingSocket - An existing WebSocket to use.
+     */
+    constructor(existingSocket: WebSocket);
```
lukehb, line 22: Cool, I didn't realise this was valid ts! I thought you could only ever have one constuctor in ts.

Frontend/implementations/typescript/src/player.ts
```diff
@@ -17,7 +17,7 @@ document.body.onload = function() {
     Logger.InitLogging(LogLevel.Warning, true);
 
 	// Create a config object
-	const config = new Config({ useUrlParams: true });
+	const config = new Config({ useUrlParams: true, webSocketProtocols: ["soap","xamp"] });
```
lukehb: Is this necessary? I would prefer to keep it exactly as it was (for compatibility reasons) for the reference implementation in case people built around it being a certain way in the past.

mcottontensor: No this was an accidental commit of testing code.

lukehb (commented):
I left a few comments and one question - just wanted to ensure we are being as safe as we can be with any changes here.

Frontend/implementations/typescript/src/player.ts
lukehb: Okay, let's fix that up and I will approve.

Common/src/Transport/WebSocketTransportNJS.ts
mcottontensor: My understanding is that it's basically just hints to typescript of how to use the one actual implementation of the constructor.

lukehb (approved):

PR #525, opened 2025-03-04
Fixing mediasoup-sdp-bridge reference in SFU

mcottontensor:
## Relevant components:
- [ ] Signalling server
- [ ] Common library
- [ ] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [x] SFU

## Problem statement:
The dependency in the SFU is incorrectly named.

## Solution
Updated the name to properly reference the package name.
Also ran the js code through the formatter.

### DIFF
--- a/SFU/package.json
+++ b/SFU/package.json
@@ -18,7 +18,7 @@
     },
     "dependencies": {
         "mediasoup_prebuilt": "^3.8.4",
-        "mediasoup-sdp-bridge": "*",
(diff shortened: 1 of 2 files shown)


lukehb:
@mcottontensor Should we backport to 5.4 and 5.3 too? Or is it not relevant there?

lukehb (approved):

mcottontensor:
It looks like it is only an issue because of the NPM workspace changes. Previous references were more direct.

PR #532, opened 2025-03-05
Mediasoup update

mcottontensor:
## Relevant components:
- [ ] Signalling server
- [ ] Common library
- [ ] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [x] SFU

## Problem statement:
We were previously distributing the binaries for the mediasoup workers ourselves but latest versions of mediasoup now provide the binaries themselves.

## Solution
Changed the dependency to mediasoup proper and updated a few pieces of code where the library had changed behaviour.
Have also updated the mediasoup-sdp-bridge package so that it's no longer a prebuilt package and is now built from source that is forked from the original source. This allows history tracking and easier development in the future.


SFU/sfu_server.js
```diff
@@ -298,9 +298,6 @@ async function setupPeerDataChannels(peerId) {
     // streamer data consumer (consumes peer data)
     peer.streamerDataConsumer = await streamer.transport.consumeData({ dataProducerId: peer.peerDataProducer.id });
 
-    streamer.transport._sctpStreamIds[nextStreamerSCTPStreamId] = 1;
-    streamer.transport._sctpStreamIds[nextPeerSCTPStreamId] = 1;
-
```
lukehb, line 303: I assume these lines were significant? Are they correctly set to the appropriate values in the new version?

lukehb (commented):
Just want to sanity check two lines that were removed.

SFU/sfu_server.js
mcottontensor: So that property has been locked down by mediasoup so we cant get to it without hacky bs. I went looking through the code to see if the behaviour was auto handled or something now but couldn't nail it down completely. So I just removed it and tried the SFU with multiple connections. I was able to control the stream on both connections and I noticed the output showing that the stream id values were increasing. So it appears to not be needed.

mcottontensor:
I just now removed the hacky, pre-built package mediasoup-sdp-bridge in favor of an actual published package from my own repo. We can probably move it to the TW group.
I did this because it allows us to see the history of changes we have done to the base project and also provides the type info with the package.
The original repo is just a placeholder for a theoretical solution from mediasoup themselves, but hasn't been touched in 5 years.
OpenVidu created a fork and suggested a solution which we adapted for our own uses a few years ago. The PR was eventually abandoned but the fork continues to be worked on. So I have forked their fork and reintroduced the changes we originally did, but this time I've implemented them in the typescript.
This should be much more maintainable and have much more visibility into where everything came from.

New NPM package was published @ https://www.npmjs.com/package/@epicgames-ps/mediasoup-sdp-bridge

SFU/sfu_server.js
lukehb: Sounds good them

mcottontensor:
Alright, I've added the mediasoup-sdp-bridge package as a git subtree now. This means that its basically a clone of the repo but committed into this repo. So we can make commits on top of it, but it's also easy to merge in upstream changes should we choose to.
This should be a decent solution when we can't make a new fork.

lukehb (approved):

PR #565, opened 2025-04-03
Action audit

mcottontensor:
Going over the actions and doing a few things.
- Removing duplicated push and pull_request triggers
- Removing multiple install commands when one root will work.
- Removing references to UE5.5 in favor of wildcard matching


.github/workflows/create-gh-release.yml
```diff
@@ -47,31 +49,37 @@ jobs:
...
       with:
         node-version: ${{ steps.get_node_version.outputs.node_version }}
         registry-url: 'https://registry.npmjs.org'
 
+    - name: Install dependencies.
+      run: npm install
+
     - name: Build common library
       working-directory: ./Common
-      run: npm install && npm run build
```
lukehb, line 59: Good point!

.github/workflows/healthcheck-frontend.yml
```diff
@@ -2,14 +2,6 @@ name: Check health of frontend
...
 on:
   workflow_dispatch:
-  push:
-    paths:
-      - ".github/workflows/healthcheck-frontend.yml"
-      - "Common/**"
-      - "Frontend/**"
-      - "Signalling/**"
-      - "SignallingWebServer/**"
-      - "Extras/**"
```
lukehb, line 12: Okay, so we are saying only do healthcheck on PR now. I think that makes sense.

mcottontensor: Yeah. Generally the only way things should get into the repo is via PR and we should only merge the PR if the actions pass. So there should be no reason to do it on push.

lukehb (approved):
Looks like a good change to me.

One unrelated thing this reminded me of is I think we could optimise `start.bat --build` time by only building esm instead of both cjs and esm. As in that case, you only care about getting something up and running, if you somehow broke cjs and not esm the PR would catch it anyway.

PR #567, opened 2025-04-03
Platform scripts audit

mcottontensor:
## Relevant components:
- [ ] Signalling server
- [ ] Common library
- [ ] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [x] Platform scripts
- [ ] SFU

## Problem statement:
The platform scripts have been problematic for a while, causing multiple builds of the same package, slow startup and failed startups.

## Solution
Have gone over the windows and linux scripts and made sure they are focused on what needs to be built and only building that. The usage text has also been updated to properly reflect the options.


SignallingWebServer/platform_scripts/cmd/common.bat
```diff
@@ -17,28 +17,35 @@ echo.
 echo    Usage:
 echo        %0 [--help] [--publicip ^<IP Address^>] [--turn ^<turn server^>] [--stun ^<stun server^>] [server options...]
 echo    Where:
-echo        --help                  Print this message and stop this script.
-echo        --publicip ^<ip^>         Define public ip address (using default port) for turn server
```
lukehb, line 21: Why remove the parameters being listed?

lukehb (commented):
~~I am going to give this a local test on my workspace, I'll report back.~~ It works well! I have some left some feedback below on a fix things to consider.

Frontend/implementations/typescript/package.json
```diff
@@ -9,6 +9,7 @@
         "build": "npx webpack --config webpack.prod.js",
         "build:dev": "npx webpack --config webpack.dev.js",
         "build:esm": "npx webpack --config webpack.esmodule.js",
+        "build:cjs": "npx webpack --config webpack.common.js",
```
lukehb: This a funny one actually on closer inspection.

We have the following webpack build configs for the reference frontend:

`webpack.common.js` - This is the base config used by both `webpack.dev.js` and `webpack.prod.js` - but perhaps confusingly named is not actually configuring webpack to build `commonjs`, but rather `umd` (which is correct considering you can't run cjs in the browser and webpack transforms the `cjs` deps into browser compatible `umd` for us). One solution to avoid this confusion is to rename this file `webpack.base.js`. Of additional note, because the `webpack.common.js` configuration does not specify a mode (`development` or `production`) webpack emits some warnings saying falling back to development etc, so this config should probably never be called directly.

`webpack.dev.js` - The dev configuration of the `umd` build (with inline source maps). This is probably the one we want the `start.bat` script to use.

`webpack.prod.js` - The prod configuration of the `umd` build that does minification etc so you end up with a smaller bundle.

`webpack.esmodule.js` - A completely different configuration that bundles into an `esm` module. Which could also be interesting for some usecases (like if someone is an all ESM codebase), but probably `umd` is the one we and most users will want to use.

```suggestion
        "build:prod": "npx webpack --config webpack.prod.js",
```

SignallingWebServer/platform_scripts/bash/common.sh
```diff
@@ -204,35 +233,25 @@ function setup_frontend() {
...
 		popd > /dev/null
 		pushd "${SCRIPT_DIR}/../../../Frontend/ui-library" > /dev/null
-		"${SCRIPT_DIR}/node/bin/npm" install
-		"${SCRIPT_DIR}/node/bin/npm" run build
+		"${NPM}" run build:cjs
 		popd > /dev/null
 		pushd "${SCRIPT_DIR}/../../../Frontend/implementations/typescript" > /dev/null
-		"${SCRIPT_DIR}/node/bin/npm" install
-		"${SCRIPT_DIR}/node/bin/npm" run build
+		"${NPM}" run build:cjs
```
lukehb:
```suggestion
		"${NPM}" run build:dev
```

SignallingWebServer/platform_scripts/cmd/common.bat
```diff
@@ -155,38 +206,24 @@ set FRONTEND_DIR=%RETVAL%
...
-    call "%SCRIPT_DIR%node\npm" install
-    call "%SCRIPT_DIR%node\npm" run build
+    call %NPM% run build:cjs
     popd
-    echo Building Epic Games reference frontend...
-    echo ----------------------------
     pushd %CD%\Frontend\implementations\typescript
-    call "%SCRIPT_DIR%node\npm" install
-    call "%SCRIPT_DIR%node\npm" run build
+    call %NPM% run build:cjs
```
lukehb:
```suggestion
    call %NPM% run build:dev
```

lukehb:
As an aside, I found when testing this that webpack building `umd` reference frontend it will selectively use either the `/dist/cjs` or `dist/esm` output of each of the packages depending which one exists on disk, with a preference for `esm`.

In terms of the final bundle that is produced in these two outcomes (deps being either all cjs or all esm) the bundle size is very similar.

Building the dev bundle is about 2.4mb on disk vs prod which is 370kb. (As an additional note when I reviewed the GH releases our bundle size does seem to have grown by about 100kb since October last year for no reason? - Maybe we should use: https://www.npmjs.com/package/webpack-bundle-analyzer )

The reason I mention the bundle size is it fine for `start.bat` to use larger dev bundle imo, but the GH release ships should prod bundle (which it does not, but just mentioning this in case you make more changes to this PR).

mcottontensor:
> The reason I mention the bundle size is it fine for `start.bat` to use larger dev bundle imo, but the GH release ships should prod bundle (which it does not, but just mentioning this in case you make more changes to this PR).

Looking at the release action it uses the `build` npm script which calls webpack with the prod configuration?

SignallingWebServer/platform_scripts/cmd/common.bat
mcottontensor: It was just a copy/paste from the bash help which was more up to date. We can put it back but I feel its self explanatory.

lukehb:
> > The reason I mention the bundle size is it fine for `start.bat` to use larger dev bundle imo, but the GH release ships should prod bundle (which it does not, but just mentioning this in case you make more changes to this PR).
> 
> Looking at the release action it uses the `build` npm script which calls webpack with the prod configuration?

Quite right! Just ignore that bit then. I still think the growth in bundle size is interesting and should be looked into in a new ticket/PR. But is certainly not a blocker for this one.

lukehb (changes requested):
Some rfc

SignallingWebServer/platform_scripts/cmd/common.bat
lukehb: Yeah, fair enough.

PR #601, opened 2025-04-27
Fixing dependency issues that arise in the docker image.

mcottontensor:
## Relevant components:
- [x] Signalling server
- [ ] Common library
- [ ] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
Occasionally when building the signalling docker image, package issues could randomly show up.

## Solution
This was simply caused by the fact that the package-lock file was not copied into the image. This change copies that file in which locks the packages to known working configurations.

### DIFF
--- a/SignallingWebServer/Dockerfile
+++ b/SignallingWebServer/Dockerfile
@@ -4,6 +4,7 @@ FROM node:lts
 WORKDIR /SignallingWebServer
 COPY NODE_VERSION ./NODE_VERSION
 COPY package.json ./package.json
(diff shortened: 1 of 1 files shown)


lukehb:
@mcottontensor Look's good to me. Does this need a changeset? Wdyt?

lukehb (approved):

mcottontensor:
Nah no real change just a fixed config.

PR #607, opened 2025-05-05
Better handling of node in Linux scripts.

mcottontensor:
## Relevant components:
- [ ] Signalling server
- [ ] Common library
- [ ] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [x] Platform scripts
- [ ] SFU

## Problem statement:
A few issues:
- The handling of node sometimes causes issues on some environments.
- The semver comparison function does not work on mac.

## Solution
For the node installations, we now just check whatever node is installed. If it's greater or equal to our requirements we do nothing. Otherwise we install node in our local location. If we installed node, then we update PATH to point to the new location.
If node_modules doesnt exist or package-lock is newer than the directory, we install dependencies, otherwise we assume they're up to date.
Additionally I've also added a printout of the node and npm versions used in the script.
I've removed the weird NPM variable that was used to point to the specific npm version. Since we now properly update the path it should not be needed. Also, the way we call sudo now should also allow the correct path variable to carry over to the new environment.
The version check function also now uses a simplified version utilizing 'sort -V' which is designed to sort version numbers. This should work across OSs like Linux and Mac.

### DIFF
--- a/.backportrc.json
+++ b/.backportrc.json
@@ -1,7 +1,7 @@
 {
-  "repoOwner": "EpicGamesExt",
-  "repoName": "PixelStreamingInfrastructure",
-  "targetBranchChoices": ["master", "UE5.2", "UE5.3", "UE5.4", "UE5.5"],
-  "autoMerge": true,
-  "autoMergeMethod": "squash"
+    "repoOwner": "EpicGamesExt",
+    "repoName": "PixelStreamingInfrastructure",
+    "targetBranchChoices": ["master", "UE5.2", "UE5.3", "UE5.4", "UE5.5", "LatencyTest"],
+    "autoMerge": true,
+    "autoMergeMethod": "squash"
 }
--- a/SignallingWebServer/platform_scripts/bash/common.sh
+++ b/SignallingWebServer/platform_scripts/bash/common.sh
@@ -168,7 +168,6 @@ function setup_node() {
         node_url="https://nodejs.org/dist/$NODE_VERSION/node-$NODE_VERSION-linux-x64.tar.gz"
     fi
 
-    node_version="101.1.1" #TEMP debug testing
     check_and_install "node" "$node_version" "$NODE_VERSION" "curl $node_url --output node.tar.xz
                                                                 && tar -xf node.tar.xz
                                                                 && rm node.tar.xz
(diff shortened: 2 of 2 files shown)


.backportrc.json
mcottontensor, line 1: This is just piggy backing with this change. I figure it's useful to backport to LatencyTest.

SignallingWebServer/platform_scripts/bash/common.sh
```diff
@@ -372,27 +360,26 @@ function start_process() {
...
-# NPM = The npm command path
 # SERVER_ARGS The arguments to be passed to the server
 function start_wilbur() {
     pushd "${SCRIPT_DIR}/../../../SignallingWebServer" > /dev/null
 
     echo "Starting wilbur signalling server use ctrl-c to exit"
     echo "----------------------------------"
     
-    start_process "sudo PATH=\"$PATH\" \"$NPM\" start -- ${SERVER_ARGS}"
+    echo "sudo env \"PATH=$PATH\" npm start -- ${SERVER_ARGS}"
```
lukehb: Is the `echo` here just for debug?

SignallingWebServer/platform_scripts/bash/common.sh
```diff
@@ -169,14 +149,11 @@ function setup_node() {
     pushd "${SCRIPT_DIR}/../../.." > /dev/null
 
     # setup the path so we're using our node. required when calling npm
-    PATH="${SCRIPT_DIR}/node/bin:$PATH"
+    # PATH="${SCRIPT_DIR}/node/bin:$PATH"
```
lukehb: These comments can go (or be moved to line 177)

SignallingWebServer/platform_scripts/bash/common.sh
mcottontensor: Yep. Will remove.

lukehb (approved):
Looks like a good change to me. CI is passing from what I can see. I left a few minor comments/questions - but those are not merge blockers.

mcottontensor:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|UE5.5|[](https://github.com/EpicGamesExt/PixelStreamingInfrastructure/pull/610)|

### Questions ?

PR #617, opened 2025-05-15
Merging 5.6 updates into master

mcottontensor:
Branch 5.6 has now been cut and tests are passing.
This PR will bring master up to 5.6 as the current dev stream.


.github/workflows/healthcheck-libs-with-public-deps.yml
```diff
@@ -12,8 +12,20 @@ env:
 jobs:
   build-using-public-deps:
     runs-on: ubuntu-latest
-    outputs:
-      continue: ${{ steps.set_output.outputs.continue }}
+    strategy:
+      matrix:
+        package: 
+          - path: Common
+            needs-lock: true
```
lukehb: What's `needs-lock`?

lukehb: Seems it indicates if package-lock should be copied?

lukehb (approved):
Okay, looks good! Nice work getting everything green, I also manually tested some of the releases, things seem in working order to me!

.github/workflows/healthcheck-libs-with-public-deps.yml
mcottontensor: Yep. When we delete the root package.json file to fore npm to use published libs, we also lose the locked package versioning. Theres currently a package we use that if unlocked it misbehaves and kills the build. So when building our libraries we should be keeping the lockfile.
However for some reason, the typescript implementation didnt play nice when using the lockfile. I think it's related to the "*" version requirement in the dependencies. So i had to add a flag to only conditionally copy the lockfile.

PR #653, opened 2025-05-29
Fixing argument issues with Wilbur and platform scripts

### LINKED ISSUE #650: [BUG] - Signalling Server Image start.sh doesn't allow to provide your own peer connection options
**UE Version:**
UE 5.5.4

**Frontend Version:**
>=UE5.5

**Problem component**
Signalling Server container image

**Description**
The start.sh will always call "setup_turn_stun" which will set the TURN_SERVER and STUN_SERVER variables.
https://github.com/EpicGamesExt/PixelStreamingInfrastructure/blob/master/SignallingWebServer/platform_scripts/bash/start.sh#L10

Therefore, both variables will always be set when the start.sh will reach here: https://github.com/EpicGamesExt/PixelStreamingInfrastructure/blob/master/SignallingWebServer/platform_scripts/bash/start.sh#L13
There is no way to avoid the two variables (TURN_SERVER and STUN_SERVER) being set.

This always adds the --peer_options to the SERVER_ARGS variable.

Server command line arguments:  --peer_options={ "iceServers":[ { "urls": ["turn:10.10.10.10:3478"], "username": "user", "credential": "password" } ], "iceTransportPolicy": "relay" } --serve --https_redirect --console_messages verbose --log_config --public_ip=212.23.236.67 --peer_options='{"iceServers":[{"urls":["stun:stun.l.google.com:19302","turn:212.23.236.67:19303"],"username":"PixelStreamingUser","credential":"AnotherTURNintheroad"}]}' --http_root='/SignallingWebServer/SignallingWebServer/platform_scripts/bash/../../www'

This leads to a second --peer_options, although already provided.

Also the provided --peer_options is not correctly forwarded to the process, as it looses it's leading and tailing apostrophs, leading to wrong parsing.

> node ./dist/index.js --serve --console_messages verbose --log_config --https_redirect --peer_options={ iceServers:[ { urls: [turn:10.10.10.10:3478], username: user, credential: password } ], iceTransportPolicy: relay } --serve --https_redirect --console_messages verbose --log_config --public_ip=212.23.236.67 --peer_options={"iceServers":[{"urls":["stun:stun.l.google.com:19302","turn:212.23.236.67:19303"],"username":"PixelStreamingUser","credential":"AnotherTURNintheroad"}]} --http_root=/SignallingWebServer/SignallingWebServer/platform_scripts/bash/../../www

SyntaxError: Expected property name or '}' in JSON at position 1 (line 1 column 2)

**Steps to Reproduce:**
docker run pixelstreamingunofficial/pixel-streaming-signalling-server:5.5 --peer_options='{ "iceServers":[ { "urls": ["turn:10.10.10.10:3478"], "username": "user", "credential": "password" } ], "iceTransportPolicy": "relay" }'

**Expected behavior**
Server command line arguments:  --peer_options='{ "iceServers":[ { "urls": ["turn:10.10.10.10:3478"], "username": "user", "credential": "password" } ], "iceTransportPolicy": "relay" }' --serve --https_redirect --console_messages verbose --log_config --public_ip=212.23.236.67  --http_root='/SignallingWebServer/SignallingWebServer/platform_scripts/bash/../../www'

**Screenshots**

 - OS: Linux

mcottontensor:
## Relevant components:
- [x] Signalling server
- [ ] Common library
- [ ] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [x] Platform scripts
- [ ] SFU

## Problem statement:
The current platform scripts don't allow you to pass --peer_options (and possibly some other parameters) to Wilbur because the argument is constructed in the platform scripts themselves and replaces any user passed peer_options.

## Solution
The platform scripts now check if any parameter it would normally pass to Wilbur already were passed by the user. If so it skips adding that parameter to the call to Wilbur.

Another change that comes with this is the implementation of an explicit delimeter between script arguments and Wilbur arguments that will be passed directly. Now anything before the `--` marker is treated as a script parameter, and if an unknown parameter is found an error is produced (previously unknown parameters would be passed to Wilbur). And anything after the marker is passed directly to Wilbur. This means after this update, users will need to properly separate script parameters and Wilbur parameters with `--`.

One final change with this PR is that there is a new `--peer_options_file` parameter for Wilbur. JSON is notoriously bad to pass as a command line argument value due to formatting, escaping, spaces etc. It does not parse well on the command line. To aleviate this, I have added the new `--peer_options_file` parameter that allows users to provide a JSON file for the peer options. This file is read, parsed and passed in as the peer options for the signalling server..

Fixes #650

### DIFF
--- a/SignallingWebServer/src/index.ts
+++ b/SignallingWebServer/src/index.ts
@@ -147,6 +147,12 @@ program
             .argParser(JSON.parse)
             .default(config_file.peer_options || '')
     )
+    .addOption(
+        new Option(
+            '--peer_options_file <filename>',
+            'Additional JSON data to send in peerConnectionOptions of the config message. This allows you to provide JSON data without having to deal with it on the command line.'
+        ).default(config_file.peer_options_file || '')
+    )
     .option(
         '--reverse-proxy',
         'Enables reverse proxy mode. This will trust the X-Forwarded-For header.',
@@ -199,6 +205,20 @@ InitLogging({
     logLevelConsole: options.log_level_console,
     logLevelFile: options.log_level_file
 });
+
+// read the peer_options_file
+if (options.peer_options_file) {
+    if (!fs.existsSync(options.peer_options_file)) {
+        Logger.error(`peer_options_file "${options.peer_options_file}" does not exist.`);
+        throw Error('Failed.');
+    }
+
+    options.peer_options = JSON.parse(fs.readFileSync(options.peer_options_file, 'utf-8'));
+} else if (options.peer_options) {
+    Logger.warn(
+        `The --peer_options cli flag has many issues with passing JSON data on the command line. It is recommended that you use --peer_options_file instead.`
+    );
+}
 
 // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
 Logger.info(`${pjson.name} v${pjson.version} starting...`);
--- a/SignallingWebServer/src/index.ts
+++ b/SignallingWebServer/src/index.ts
@@ -205,7 +205,7 @@ Logger.info(`${pjson.name} v${pjson.version} starting...`);
 if (options.log_config) {
     Logger.info('Config:');
     for (const key in options) {
-        Logger.info(`"${key}": "${options[key]}"`);
+        Logger.info(`"${key}": ${JSON.stringify(options[key])}`);
     }
 }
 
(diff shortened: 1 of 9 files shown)


SignallingWebServer/src/index.ts
```diff
@@ -200,12 +206,26 @@ InitLogging({
     logLevelFile: options.log_level_file
 });
 
+// read the peer_options_file
+if (options.peer_options_file) {
+    if (!fs.existsSync(options.peer_options_file)) {
+        Logger.error(`peer_options_file "${options.peer_options_file}" does not exist.`);
+        throw Error('Failed.');
```
lukehb:
```suggestion
        throw Error(`Failed to find a peer options config file a file called ${options.peer_options_file}.`);
```

lukehb (changes requested):
Slight change to improve error thrown

PR #673, opened 2025-06-24
ping players every <Interval>, receive pong response

mnecci-DQ:
## Relevant components:
- [x] Signalling server
- [ ] Common library
- [ ] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
I was encountering a problem where my frontend client was loosing the webrtc connection at apparently random times. After some investigation I discovered the following:
- disconnection occurred evry 5 minutes exactly
- this was not due to any errors on the client
- this was not due to any errors on UE
- this was not due to any errors on the signalling server
- the AFK flag was disabled on the client
- 300 seconds (5 minutes) is exactly the timeout time of an idle websocket connection
- when the websocket connection to the signalling server is lost, the webrtc connection is also lost (this is also used by in the repo as a reset connection mechanism)
- the streamer and signalling server periodically exchange ping messages
- the PixelStreaming object is designed to respond to ping message from the signalling server with a pong messages
- there's no logic in cirrus about sending ping messages to the client

## Solution
This PR adds a piece of logic that sends a ping message every [INTERVAL] and expects a pong message in return, this fixes my client-side periodic disconnection problem 

I tested this in my dev environment, unfortunately I did not add any tests. The logic is simple enough and I already implemented it in my fork.

### DIFF
--- a/SignallingWebServer/cirrus.js
+++ b/SignallingWebServer/cirrus.js
@@ -748,7 +748,7 @@ function onSFUMessageStartStreaming(sfuPlayer, msg) {
 function onSFUMessageStopStreaming(sfuPlayer, msg) {
 	const sfuStreamerComponent = sfuPlayer.getSFUStreamerComponent();
 	logIncoming(sfuStreamerComponent.id, msg);
-if (!streamers.has(sfuStreamerComponent.id)) {
+	if (!streamers.has(sfuStreamerComponent.id)) {
 		console.error(`SFU ${sfuStreamerComponent.id} is not registered as a streamer or streaming.`)
 		return;
 	}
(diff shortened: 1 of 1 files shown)


mnecci-DQ:
I realized this was breaking other functionalities, I still believe this is a needed feature, but it needs more testing

mnecci-DQ:
after further testing I realized it was all working fine, I was just very tired.

mcottontensor:
I'll be accepting this specifically for the 5.4 release as in that release there is no ping between players and the signalling server. However from 5.5 onwards there is a ping from the player frontend to the signalling server so this change is not needed.

PR #679, opened 2025-07-17
Fixing Unsafe AFKOverlay constructed from library input 

odaysec:
https://github.com/EpicGamesExt/PixelStreamingInfrastructure/blob/28e89962b90f69dfd9ac4bbbf703df8631e022c1/Frontend/ui-library/src/Overlay/AFKOverlay.ts#L47-L47

fix the issue, we need to ensure that the `countdown` parameter is safely handled before being inserted into the DOM. The best approach is to avoid using `innerHTML` for dynamic content and instead use safer alternatives like `textContent` or DOM manipulation methods. This ensures that no untrusted HTML or JavaScript can be executed.

**Steps to fix:**
- Replace the use of `innerHTML` with DOM manipulation methods to construct the HTML structure dynamically.
- Ensure that the `countdown` value is safely inserted as plain text using `textContent`.

**Changes required:**
- Modify the `updateCountdown` method to use `document.createElement` and `textContent` instead of `innerHTML`.

---

When a library function dynamically constructs HTML in a potentially unsafe way, then it's important to document to clients of the library that the function should only be used with trusted inputs. If the function is not documented as being potentially unsafe, then a client may inadvertently use inputs containing unsafe HTML fragments, and thereby leave the client vulnerable to cross-site scripting attacks.

The following has a library function that renders a boldface name by writing to the innerHTML property of an element.
```ts
module.exports = function showBoldName(name) {
  document.getElementById('name').innerHTML = "<b>" + name + "</b>";
}
```
This library function, however, does not escape unsafe HTML, and a client that calls the function with user-supplied input may be vulnerable to cross-site scripting attacks. The library could either document that this function should not be used with unsafe inputs, or use safe APIs such as `innerText`.
```ts
module.exports = function showBoldName(name) {
  const bold = document.createElement('b');
  bold.innerText = name;
  document.getElementById('name').appendChild(bold);
}
```

#### References
- [DOM based XSS Prevention Cheat Sheet](https://www.owasp.org/index.php/DOM_based_XSS_Prevention_Cheat_Sheet)
- [XSS (Cross Site Scripting) Prevention Cheat Sheet](https://www.owasp.org/index.php/XSS_%28Cross_Site_Scripting%29_Prevention_Cheat_Sheet)
- [DOM Based XSS](https://www.owasp.org/index.php/DOM_Based_XSS)
- [Types of Cross-Site Scripting](https://www.owasp.org/index.php/Types_of_Cross-Site_Scripting)
[CWE-79](https://cwe.mitre.org/data/definitions/79.html)
[CWE-116](https://cwe.mitre.org/data/definitions/116.html)

## Relevant components:
- [x] Signalling server
- [ ] Common library
- [x] Frontend library
- [x] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:

## Solution

## Documentation

### DIFF
--- a/Frontend/ui-library/src/Overlay/AFKOverlay.ts
+++ b/Frontend/ui-library/src/Overlay/AFKOverlay.ts
@@ -44,6 +44,18 @@ export class AFKOverlay extends ActionOverlay {
      * @param countdown - the count down number to be inserted into the span for updating
      */
     public updateCountdown(countdown: number): void {
-        this.textElement.innerHTML = `<center>No activity detected<br>Disconnecting in <span id="afkCountDownNumber">${countdown}</span> seconds<br>Click to continue<br></center>`;
+        const centerElement = document.createElement('center');
+        centerElement.appendChild(document.createTextNode('No activity detected'));
+        centerElement.appendChild(document.createElement('br'));
+        centerElement.appendChild(document.createTextNode('Disconnecting in '));
+        const countdownSpan = document.createElement('span');
+        countdownSpan.id = 'afkCountDownNumber';
+        countdownSpan.textContent = countdown.toString();
+        centerElement.appendChild(countdownSpan);
+        centerElement.appendChild(document.createTextNode(' seconds'));
+        centerElement.appendChild(document.createElement('br'));
+        centerElement.appendChild(document.createTextNode('Click to continue'));
+        this.textElement.innerHTML = ''; // Clear existing content
+        this.textElement.appendChild(centerElement);
     }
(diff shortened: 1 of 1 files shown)


mcottontensor:
While this is a good change from a security viewpoint, this behavior in the code is not the only location for it. There are multiple places where innerHtml is used as this kind of security issue was not a focus when developed.

Fixing one location seems arbitrary and we would much rather accept a PR that addresses all locations.

PR #694, opened 2025-09-11
Fix Pixel Streaming intermittently fails to make a network connection even with correct ICE candidates

### LINKED ISSUE #692: [BUG] Intermittent ICE candidate failure when correct candidates are present
**UE Version:**
5.6 (and all prior versions)

**Frontend Version:**
5.6 (and all prior versions)

**Problem component**
Frontend

**Description**
In some instances WebRTC will fail to make a network connection even if the relevant ICE candidates are present. This behaviour manifests with certain connections (often Windows UE running on a slow machine where `srflx` is required on both sides) and if the connection is repeatedly retried it will intermittently connect.

**Proposed fix**

Pixel Streaming should always connect if it has the relevant ICE candidates.

lukehb:
## Relevant components:
- [x] Frontend library

## Problem statement:
In some instances Pixel Streaming/WebRTC will fail to make a network connection even if the relevant ICE candidates are present. This behaviour manifests with certain connections (often Windows UE running on a slow machine where `srflx` is required on both sides) and if the connection is repeatedly retried it will intermittently connect. See: https://github.com/EpicGamesExt/PixelStreamingInfrastructure/issues/692

## Solution
Through investigation it was found that:

-  The Pixel Streaming plugin (in all versions) will send ICE candidates with `sdpMid` and `sdpMLineIndex`
- `sdpMid` and `sdpMLineIndex` are not required if using `bundle` mode (where media is multiplexed over one transport)
- `bundle` is Pixel Streaming's default (and only supported mode) for WebRTC connections
- `sdpMid` can be set to an empty string `""` and `sdpMLineIndex` can be omitted and the connection can proceed normally

While the WebRTC spec says candidates can be trickled in any order, we observed that if a candidate containing `sdpMid:n` (where n is any value other than the master bundle, e.g. `0` in our case) was received before the candidate containing `sdpMid:m` (where m is the master bundle, e.g. `0`) then the connection would fail. The fix was simply to remove this unnecessary information and the connection proceeds as it should and is able to connect.

## Documentation
N/A - though it should be noted with this change we explicitly are making the assumption now that Pixel Streaming must use `bundle` - which has always been an implicit assumption anyway.

I tested this PR on the following browsers/devices:

Chrome + Windows
Firefox + Windows
Chrome + iPhone
Chrome + Mac
Chrome + iPad
Safari + iPhone
Safari + Mac
Safari + iPad
Chrome + Android
Android Browser + Android

### DIFF
--- a/Frontend/library/src/WebRtcPlayer/WebRtcPlayerController.ts
+++ b/Frontend/library/src/WebRtcPlayer/WebRtcPlayerController.ts
@@ -1106,19 +1106,35 @@ export class WebRtcPlayerController {
             this.pixelStreaming._onLatencyCalculated(latencyInfo);
         };
 
-        /* When the Peer Connection wants to send an offer have it handled */
+        /* When our PeerConnection sends want to send an offer call our handler */
         this.peerConnectionController.onSendWebRTCOffer = (offer: RTCSessionDescriptionInit) => {
             this.handleSendWebRTCOffer(offer);
         };
 
-        /* Set event handler for when local answer description is set */
-        this.peerConnectionController.onSetLocalDescription = (answer: RTCSessionDescriptionInit) => {
-            this.handleSendWebRTCAnswer(answer);
+        /* Set event handler for when local description is set */
+        this.peerConnectionController.onSetLocalDescription = (sdp: RTCSessionDescriptionInit) => {
+            if (sdp.type === 'offer') {
+                this.handleSendWebRTCOffer(sdp);
+            } else if (sdp.type === 'answer') {
+                this.handleSendWebRTCAnswer(sdp);
+            } else {
+                Logger.Error(
+                    `PeerConnectionController onSetLocalDescription was called with unexpected type ${sdp.type}`
+                );
+            }
         };
 
-        /* Set event handler for when remote offer description is set */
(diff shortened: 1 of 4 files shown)


Frontend/library/src/WebRtcPlayer/WebRtcPlayerController.ts
```diff
@@ -1463,23 +1479,31 @@ export class WebRtcPlayerController {
...
+        Logger.Info(`Remote ICE candidate information received: ${JSON.stringify(iceCandidateInit)}`);
+
+        // We are using "bundle" policy for media lines so we remove the sdpMid and sdpMLineIndex attributes
+        // from ICE candidates as these are legacy attributes for when bundle is not used.
+        // If we don't do this the browser may be unable to form a media connection
+        // because some browsers are brittle if the bundle master (e.g. commonly mid=0) doesn't get a candidate first.
+        const remoteIceCandidate = new RTCIceCandidate({
+            candidate: iceCandidateInit.candidate,
+            sdpMid: ''
+        });
```
lukehb, line 1495: These are the important lines of this fix

Belchy06 (approved):
LGTM!

PR #709, opened 2025-10-02
Fix: Firefox connectivity issues caused by PR694

lukehb:
## Relevant components:
- [x] Frontend library

## Problem statement:
As a result of https://github.com/EpicGamesExt/PixelStreamingInfrastructure/pull/694 sometimes Firefox was unable to connect. It would error with `uncaught (in promise) DOMException: Cannot set ICE candidate for level=<Nothing> mid=: No such transceiver.`

## Solution
This PR patches the previous fix by hard coding the `sdpMLineIndex` to `0` which resolves the issue on Firefox and maintains working connectivity on the other browsers. This change means we assume:

1. Bundle is being used for WebRTC media
2. The master media line of the bundle is in index 0.

Both of these assumptions hold for all versions of Pixel Streaming 1 and 2 to date.

## Documentation
N/A - This is very implementation specific, but this PR should serve as a good record.

Tested:

Chrome + AWS 
Chrome + TailScale
Firefox + AWS
FireFox + TailScale
Safari + AWS
Safari + TailScale

All these combinations were able to stream correctly.

### DIFF
--- a/.changeset/light-games-wonder.md
+++ b/.changeset/light-games-wonder.md
@@ -1,0 +1,5 @@
+---
+'@epicgames-ps/lib-pixelstreamingfrontend-ue5.7': patch
+---
+
+Some versions of Firefox were unable to connect due the changes in PR#694 to overcome this issue and preserve the connectivity fixes from PR#694 we now assume the sdpMLineIndex is always 0 for bundle master media line. This change was tested on many browsers and restores connectivity with FireFox.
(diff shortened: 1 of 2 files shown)


lukehb:
If this turns out to be an issue because bundle master is no longer 0 we can get bundle master using something like:

`const bundleMasterMid = sdp.match(/^a=group:BUNDLE (.*)$/m)[1].split(" ")[0];`

Then we can just always add a candidate with that mid, the immediately after, add another candidate with whatever the real mid was. It is a little wasteful in terms of adding known failing ice candidates, but it should fix the issue.

PR #714, opened 2025-10-06
Update signalling_messages.proto

mehdy:
## Relevant components:
- [ ] Signalling server
- [x] Common library
- [ ] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
1. The `time` field of `ping` and `pong` messages needs to be of type `int64` to allow for milliseconds (which is how it's done in the implementation).
2. There are a few `playerId` fields missing from the proto messages that are actually used/sent in the implementation.

## Solution
1. Changed the type of `time` to `int64`.
2. Added the missing fields.

## Documentation

### DIFF
--- a/Common/protobuf/signalling_messages.proto
+++ b/Common/protobuf/signalling_messages.proto
@@ -246,7 +246,7 @@ message ping {
   // Should always be 'ping'
   string type = 1;
   // The current time
-  int32 time = 2;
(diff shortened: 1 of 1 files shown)


mehdy:
I still think this is a valid change so BUMP.

PR #728, opened 2025-11-20
Expose playerid for programmatic access

lukehb:
## Relevant components:
- [x] Signalling server
- [x] Frontend library

## Problem statement:
It is useful to be able to use the player id as a unique identifier that is common between UE side stats and frontend side stats; however, the player id is not actually exposed to TS/JS because the SS strips it out of signalling messages.

## Solution

- Make SS not strip the player id
- Expose the `playerId` as a field that is accessible

I ran this locally and was able to access the `playerId` using the `window.pixelStreaming` using my developer console in Chrome.


Belchy06 (approved):
LGTM

mcottontensor (approved):

PR #734, opened 2025-11-24
SecurityError: Failed to execute 'isSessionSupported' on 'XRSystem' when in iframe

### LINKED ISSUE #722: [BUG] - SecurityError: Failed to execute 'isSessionSupported' on 'XRSystem' when in iframe
**UE Version:**
UE5.6

**Frontend Version:**
E.g. UE5.5-1.2.4

**Problem component**
Frontend

**Description**
When loading pixel streaming in an iframe and you don't have the allow="xr-spatial-tracking" the dom will throw an Exception when trying to read the property .xr

SecurityError: Failed to execute 'isSessionSupported' on 'XRSystem': Access to the feature "xr" is disallowed by permissions policy.

**Steps to Reproduce:**
1. create an iframe an embed a pixel streaming app in it.

**Expected behavior**
We should wrap the navigator.xr.isSessionSupported() call in a try catch block so that we don't get an error if we are in an iframe.

leon:
fixes: #722

## Relevant components:
- [ ] Signalling server
- [ ] Common library
- [ ] Frontend library
- [x] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
#722 

## Solution
Wraps the check in a try catch block and returns false if it throws

It only fixes a bug, does not introduce any changes to code paths.

### DIFF
--- a/Frontend/library/src/WebXR/WebXRController.ts
+++ b/Frontend/library/src/WebXR/WebXRController.ts
@@ -491,7 +491,9 @@ export class WebXRController {
             if (navigator.xr) {
                 return navigator.xr.isSessionSupported(mode);
             }
-        } catch {}
+        } catch {
+            return Promise.resolve(false);
+        }
 
         return Promise.resolve(false);
     }
--- a/Frontend/library/src/WebXR/WebXRController.ts
+++ b/Frontend/library/src/WebXR/WebXRController.ts
@@ -485,6 +485,8 @@ export class WebXRController {
             Logger.Info('WebXR requires https, if you want WebXR use https.');
         }
 
+        // Wrap in try-catch because access to XR object can be denied due
+        // to browser security permissions (e.g. streaming from an iframe)
         try {
             if (navigator.xr) {
                 return navigator.xr.isSessionSupported(mode);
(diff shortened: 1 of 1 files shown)


Frontend/library/src/WebXR/WebXRController.ts
```diff
@@ -485,12 +485,12 @@ export class WebXRController {
...
         }
 
-        if (navigator.xr) {
-            return navigator.xr.isSessionSupported(mode);
-        } else {
-            return new Promise<boolean>(() => {
-                return false;
-            });
-        }
+        try {
```
lukehb, line 490: Adding a comment here to explain try-catch usage

```suggestion
        // Wrap in try-catch because access to XR object can be denied due
        // to browser security permissions (e.g. streaming from an iframe)
        try {
```

Frontend/library/src/WebXR/WebXRController.ts
```diff
@@ -485,12 +485,12 @@ export class WebXRController {
...
+            if (navigator.xr) {
+                return navigator.xr.isSessionSupported(mode);
+            }
+        } catch {}
```
lukehb: Our linter does not allow empty blocks, so I added resolve false to the catch block.

```suggestion
        } catch {
            return Promise.resolve(false);
        }
```

leon:
Great!

It was such a simple fix, so forgot to run linting, will do that next time. 🤓

Issue #754, opened 2025-12-19
[BUG] - Message length mismatch for non ASCII-characters


FloEdelmann:
### UE Version

UE 5.3.2.0

### Frontend Version

- `@epicgames-ps/lib-pixelstreamingcommon-ue5.5`: 0.3.2
- `@epicgames-ps/lib-pixelstreamingfrontend-ue5.5`: 1.2.5

### Problem component

Frontend

### Description

When sending a UIInteraction via `stream.emitUIInteraction({…})` containing a non-ASCII character like a German umlaut (äöüÄÖÜß), the message cannot be parsed correctly on the Unreal side. After further investigation, I noticed that this is because a mismatch of different string length calculation methods:

To determine the initial buffer size, [TextEncoder.encode](https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder/encode)'s length is used:

https://github.com/EpicGamesExt/PixelStreamingInfrastructure/blob/067138e476b706e1fc602df67bababac9e2b263d/Frontend/library/src/UeInstanceMessage/SendMessageController.ts#L91-L94

But later, when the string is written into the buffer, raw [String.length](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/length) is used:

https://github.com/EpicGamesExt/PixelStreamingInfrastructure/blob/067138e476b706e1fc602df67bababac9e2b263d/Frontend/library/src/UeInstanceMessage/SendMessageController.ts#L132-L137

For ASCII characters, this length is the same, but for e.g. umlauts, the length differs:

```js
new TextEncoder().encode('a')         // → Uint8Array [ 97 ]
new TextEncoder().encode('a').length  // → 1
'a'.charCodeAt(0)                     // → 97
'a'.length                            // → 1

new TextEncoder().encode('ü')         // → Uint8Array [ 195, 188 ]
new TextEncoder().encode('ü').length  // → 2
'ü'.charCodeAt(0)                     // → 252
'ü'.length                            // → 1
```

This means that more space is reserved in the buffer than is actually written, so trailing null bytes are sent in the message, which cause the Unreal-side C++ JSON parser to trip.

### Steps to Reproduce

Open a stream and call `stream.emitUIInteraction({ "foo": "bär" })`.

### Expected behavior

The JSON object `{ "foo": "bär" }` is deserialized correctly in Unreal.

### Suggested fix

Use raw String.length anywhere, i.e. replace

[...]

PR #758, opened 2026-01-13
Fix: incorrect string length when emitting descriptors with non-ASCII characters

Belchy06:
## Relevant components:
- [ ] Signalling server
- [ ] Common library
- [X] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
As mentioned in https://github.com/EpicGamesExt/PixelStreamingInfrastructure/issues/754, when emitting a UI interaction with non-ascii characters, the calculated `byteLength` is incorrect.

## Solution
Remove the unnecessary usage of `textEncoder` and use the direct string length.

Tested with UE 5.7 (PixelStreaming2)

### DIFF
--- a/Frontend/library/src/UeInstanceMessage/SendMessageController.ts
+++ b/Frontend/library/src/UeInstanceMessage/SendMessageController.ts
@@ -62,7 +62,6 @@ export class SendMessageController {
         }
 
         let byteLength = 0;
-        const textEncoder = new TextEncoder();
         // One loop to calculate the length in bytes of all of the provided data
         messageData.forEach((element: number | string, idx: number) => {
             const type = messageFormat.structure[idx];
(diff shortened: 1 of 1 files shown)


Frontend/library/src/UeInstanceMessage/SendMessageController.ts
```diff
@@ -91,7 +90,7 @@ export class SendMessageController {
                     // 2 bytes for string length
                     byteLength += 2;
                     // 2 bytes per characters
```
FloEdelmann:
```suggestion
                    // 2 bytes per character
```

Belchy06: Fixed. ty!

mcottontensor (approved):

Issue #768, opened 2026-01-16
[BUG] - After the signaling server and the program are successfully connected, the webpage still fails to display the screen.


17681270934:
**UE Version:**
E.g. UE 5.5.4

**Frontend Version:**
E.g. UE5.5Future versions

**Problem component**
E.g. Signalling Server, Frontend

**Description**
After the signaling server and the program are successfully connected and all settings are correct, the background screen still cannot be loaded on the web page. This issue has been present since the UE5 version. The reason is that in version 5.4, the IP address can be manually modified (specifically, the:SetPublicIP function). After the update in version 5.5, the IP address is automatically obtained from the common.bat file. This is the main cause of the problem. When there are restrictions of NAT type, and the network environment (such as enterprise networks, campus networks, or certain home routers) blocks the STUN/TURN protocols, and the firewall/UDP port is blocked, WebRTC cannot establish a connection. Even if all steps are correct, the screen still cannot be loaded. Currently, my solution is to modify the:SetPublicIP function in common.bat to manually set the IP address. This will work. I hope the team can solve this problem with a more convenient and stable technical solution. Thank you.

**Steps to Reproduce:**

**Expected behavior**
1. A clear and concise description of what you expected to happen.
2. Ideally laid out in steps.

**Screenshots**
If applicable, add 

[common.zip](https://github.com/user-attachments/files/24657799/common.zip)

<img width="2240" height="762" alt="Image" src="https://github.com/user-attachments/assets/db900fec-3400-4491-9f76-71817c68be01" />
[SignallingWebServer-log.txt](https://github.com/user-attachments/files/24657852/SignallingWebServer-log.txt)
[UE5.5.4run-log.txt](https://github.com/user-attachments/files/24657853/UE5.5.4run-log.txt)

screenshots to help explain your problem.

**Platform (please complete the following information):**
 - OS: [e.g. Windows 10 Pro]
 - Browser + version [e.g. chrome 100]

**Additional context**

mcottontensor:
This should be fixed with updates to the batch scripts. Please give it a try to confirm when you get a chance.

PR #780, opened 2026-02-02
fixing public ip stomping in windows batch script.

mcottontensor:
## Relevant components:
- [ ] Signalling server
- [ ] Common library
- [ ] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [x] Platform scripts
- [ ] SFU

## Problem statement:
When supplying --publicip on the windows batch scripts it was getting overridden.

## Solution
The SetPublicIP step now checks that no ip has been previously specified.


mcottontensor:
fixes #768

PR #955, opened 2026-08-07
Fix/windows platform script defects

ChrisMaroglou:
## Relevant components:
- [ ] Signalling server
- [ ] Common library
- [ ] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [x] Platform scripts
- [ ] SFU

## Problem statement:

Five defects in the startup scripts, four specific to Windows and one affecting both platforms. They are independent bugs that happen to cluster around TURN configuration and script entry points.

**1. `setup.bat` and `start_turn.bat` do nothing.** Both gate their entire body on `%CONTINUE%`, which is never assigned anywhere in the repository. `setup.bat` exits with no output, and `start_turn.bat` never launches a TURN server. This goes unnoticed because `start.bat` calls `:Setup` unconditionally, so the common path works and only the standalone entry points are affected.

**2. `--turn-user` and `--turn-pass` cannot be used on Windows.** Both assign the literal value `1` instead of their argument, and omit the `SHIFT` that every other value-taking flag performs. The credential is then consumed as an unrecognised argument and the script aborts with `Unknown arg`. There is no way to configure TURN credentials on Windows.

**3. TURN credential defaults are coupled to `--turn` (both platforms).** The default username and password are set inside the "TURN_SERVER is empty" branch, which fails in both directions:
- Passing `--turn-user`/`--turn-pass` *without* `--turn` leaves `TURN_SERVER`
  empty, so the branch runs and overwrites the supplied credentials with the
  public defaults.
- Passing `--turn` *alone* skips the branch, leaving both credentials empty. The
  launch guard requires them to be non-empty, so `turnserver` silently never
  starts.

**4. TURN passwords containing shell metacharacters break argument parsing on Windows.** A password containing `&`, `|` or `>` aborts the whole parse with `& was unexpected at this time`. `SHIFT` runs inside the matched block, so every comparison below it in the same pass is evaluated against the argument's *value* rather than the next flag; an unquoted `%1` then expands to something like `IF ""p@ss&word""=="--stun"`, whose unbalanced quoting lets cmd.exe re-parse the operator.

**5. The coturn `--pidfile` argument is malformed on Windows.** The value is wrapped in backticks — PowerShell escape syntax that has no meaning in a batch file — so coturn receives them as part of the path. The file is also placed at the root of `C:`, which is not writable without elevation.

## Solution

One commit per defect, each independently revertable:

- **`execute setup and start_turn bodies on Windows`** — replaces the dead
  `%CONTINUE%` guard with the `IF errorlevel 1` check `start.bat` already uses,
  so a bad argument still aborts but the normal path proceeds. Also quotes the
  `common.bat` dispatch path in `setup.bat` to match its sibling scripts (it was
  the only one unquoted, and would break under a path containing spaces).

- **`read --turn-user and --turn-pass values on Windows`** — assigns `%~2` and
  adds the missing `SHIFT`, matching how `--turn` and `--stun` are handled.

- **`apply TURN credential defaults independently of --turn`** — moves each
  default into its own emptiness check, in both `cmd/common.bat` and
  `bash/common.sh`, so the two scripts stay in step.

- **`preserve shell metacharacters in TURN credentials on Windows`** — compares
  with `%~1` instead of `%1` and assigns with the `set "VAR=value"` form.
  Credentials are expanded with `!` at the launch site for the same reason. The
  quoted-assignment form is applied to all value-taking flags, not just the TURN
  ones, because cmd.exe substitutes and parses *every* block in the loop
  including non-matching ones — a single unquoted assignment anywhere breaks the
  entire pass.

- **`correct coturn pidfile argument on Windows`** — removes the backticks and
  moves the pidfile beside the coturn install, a directory setup already writes
  to.

**One observable behaviour change**, following from defect 3: passing `--turn <server>` now leaves the default credentials in place instead of empty strings, so `peer_options` carries `PixelStreamingUser`/`AnotherTURNintheroad` rather than empty `username`/`credential` fields, and with `--start-turn` a local `turnserver` process now starts where the empty-credential launch guard previously blocked it. This is the intended correction — `--start-turn` is documented as "Will launch the turnserver process" — but it is a change for anyone passing `--turn` together with `--start-turn`. The default credential *values* are unchanged.

## Documentation

No documentation changes. The existing `--help` output in `common.bat` already describes `--turn-user` as "Sets the turn username when using a turn server" and `--turn-pass` likewise — this PR makes the implementation match the documented behaviour rather than changing it.

No changeset is included: these files live under `SignallingWebServer`, which is marked `"private": true` and is not published to NPM.

There is no existing unit test harness for the platform scripts, so verification was manual against the real `common.bat` and `common.sh` rather than copies.

**Argument matrix** — every documented flag exercised individually, plus all 13 combined, plus the `--` passthrough, plus no arguments: all parse to the correct variables. `--help` and an unknown argument were checked on all five entry points (`setup`, `start`, `start_turn`, `start_with_stun`, `start_with_turn`)

**Regression on the specific fixes:**

| Case | Before | After |
| --- | --- | --- |
| `--turn-user alice --turn-pass s3cr3t` | `Unknown arg`, exit 1 | `alice` / `s3cr3t` |
| `--turn <server>` alone | credentials empty, TURN never started | defaults applied |
| no TURN arguments | defaults | defaults (unchanged) |
| password containing `&` | `& was unexpected at this time` | preserved intact |
| password containing `\|` `>` | parse aborted | preserved intact |
| `--frontend-dir "C:\Program Files\www"` | ok | ok |
| `setup.bat` | 0 lines of output | runs setup |

**End-to-end** — `start.bat --publicip 127.0.0.1 -- --player_port 8080` brings up a working server: HTTP 200 serving the player page, streamer listener on 8888, SFU listener on 8889, and `peer_options` correctly assembled with both STUN and TURN entries.

**common.sh** — `setup_turn_stun` verified across all four credential permutations (none / credentials only / `--turn` only / both).

 **Known limitations:**
- A value containing a caret (`^`) is doubled, because cmd.exe re-escapes carets through the `%*` re-parse in the label dispatch. Unchanged by this PR.
 - A value containing an exclamation mark (`!`) loses it: the entry points run under `setlocal enabledelayedexpansion`, so `set "TURN_PASS=%~2"` drops the `!` and `pass!word` silently becomes `password`. Unchanged by this PR. Both characters are now called out in `platform_scripts/cmd/README.txt`.
 - An `&` in an *unquoted* server argument after `--` (`-- a^&b`) breaks the script, because the bare `&` splits the `%*` dispatch line before argument parsing starts. Quote the value instead (`-- --http_root="C:\a&b\www"`) and it is preserved. 
- The first two affect every value-taking flag and predate this change. Fixing the caret means restructuring the label dispatch; fixing the `!` means turning delayed expansion off around argument parsing, which needs `endlocal` and would discard everything the parse just set.

**Affected branches:** `master`, `UE5.8`, `UE5.7`, `UE5.6`, `UE5.5`. `UE5.4` and earlier used PowerShell startup scripts and are not affected. Could a maintainer  please add auto-backport plus auto-backport-to-UE5.7, auto-backport-to-UE5.6,  and auto-backport-to-UE5.5?

### DIFF
--- a/SignallingWebServer/platform_scripts/bash/common.sh
+++ b/SignallingWebServer/platform_scripts/bash/common.sh
@@ -310,7 +310,16 @@ function set_public_ip() {
 function setup_turn_stun() {
     if [[ -z "$TURN_SERVER" ]]; then
         TURN_SERVER="${PUBLIC_IP}:19303"
+    fi
+
+    # Defaults are applied independently of TURN_SERVER so that --turn-user and
+    # --turn-pass are honoured on their own, and so that supplying --turn alone
+    # still leaves credentials set for the launch guard below.
+    if [[ -z "$TURN_USER" ]]; then
         TURN_USER="PixelStreamingUser"
+    fi
+
+    if [[ -z "$TURN_PASS" ]]; then
         TURN_PASS="AnotherTURNintheroad"
     fi
 
--- a/SignallingWebServer/platform_scripts/cmd/README.txt
+++ b/SignallingWebServer/platform_scripts/cmd/README.txt
@@ -12,3 +12,8 @@ The following are provided as handy shortcuts but mostly leverage start.bat func
 Tips:
 
 - You can provide --help to start.bat to get a list of customizable arguments.
+- Values passed to these scripts cannot contain ^ or ! characters. cmd.exe
+  doubles a caret when the scripts hand their arguments to common.bat, and
+  strips an exclamation mark under delayed expansion, so a TURN password such
+  as "pass!word" arrives as "password" with no error. Choose credentials
+  without those two characters.
--- a/SignallingWebServer/platform_scripts/cmd/common.bat
+++ b/SignallingWebServer/platform_scripts/cmd/common.bat
@@ -313,7 +313,7 @@ set TURN_PROCESS=turnserver.exe
 set TURN_REALM=PixelStreaming
 rem Credentials are expanded with ! rather than % so that characters cmd.exe
 rem treats as operators (&, |, ^, >) survive intact in a TURN password.
-set TURN_ARGS=-c ..\..\..\turnserver.conf --allowed-peer-ip=%LOCAL_IP% -p %TURN_PORT% -r %TURN_REALM% -X %PUBLIC_IP% -E %LOCAL_IP% -L %LOCAL_IP% --no-cli --no-tls --no-dtls --pidfile `"C:\coturn.pid`" -f -a -v -u !TURN_USER!:!TURN_PASS!
+set TURN_ARGS=-c ..\..\..\turnserver.conf --allowed-peer-ip=%LOCAL_IP% -p %TURN_PORT% -r %TURN_REALM% -X %PUBLIC_IP% -E %LOCAL_IP% -L %LOCAL_IP% --no-cli --no-tls --no-dtls --pidfile "%SCRIPT_DIR%coturn.pid" -f -a -v -u !TURN_USER!:!TURN_PASS!
 
 if "!START_TURN!"=="1" (
     IF NOT "!TURN_SERVER!"=="" (
--- a/SignallingWebServer/platform_scripts/cmd/common.bat
+++ b/SignallingWebServer/platform_scripts/cmd/common.bat
@@ -34,7 +34,7 @@ echo        --build-wilbur      Force build of wilbur
 echo        --deps              Force reinstall of dependencies
 echo    Everything after -- is passed directly to the signalling server executable.
 IF exist "%SCRIPT_DIR%..\..\dist" (
-    pushd %SCRIPT_DIR%..\..
+    pushd "%SCRIPT_DIR%..\.."
     call %NPM% run start --- --help
     popd
 )
@@ -138,7 +138,7 @@ GOTO PostArgs
 exit /b
 
 :SetupNode
-pushd %SCRIPT_DIR%
+pushd "%SCRIPT_DIR%"
 SET NODE_NAME=node-%NODE_VERSION%-win-x64
 if exist node\ (
   echo Node directory found...skipping install.
@@ -166,7 +166,7 @@ popd
 
 if "%INSTALL_DEPS%"=="1" (
     echo Installing dependencies...
-    pushd %SCRIPT_DIR%..\..\..
+    pushd "%SCRIPT_DIR%..\..\.."
     call %NPM% install
     popd
 )
@@ -189,14 +189,14 @@ if NOT exist "%SCRIPT_DIR%..\..\..\Signalling\dist" (
 )
 
 IF "%BUILD_COMMON%"=="1" (
-    pushd %SCRIPT_DIR%..\..\..\Common
+    pushd "%SCRIPT_DIR%..\..\..\Common"
     echo Building common library
     call %NPM% run build:cjs
     popd
 )
 
 IF "%BUILD_SIGNALLING%"=="1" (
-    pushd %SCRIPT_DIR%..\..\..\Signalling
+    pushd "%SCRIPT_DIR%..\..\..\Signalling"
     echo Building signalling library
     call %NPM% run build:cjs
     popd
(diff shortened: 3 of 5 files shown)


mcottontensor:
Thanks for pointing these out — it mostly looks good to me. There are a few things I'd like to address before continuing.

**1. `call "!TURN_PROCESS!" !TURN_ARGS!` (common.bat:328).** `CALL` parses the line a second time. An `&` character in the password can divide the command and run a different command. The `bg` branch uses `start` and is safe, but `start_turn.bat` uses this branch. Fix: remove `call`. `turnserver.exe` is an executable file, and direct execution is also synchronous.

**2. The `!` character in a credential.** Delayed expansion removes it at `set "TURN_PASS=%~2"`. The password `pass!word` becomes `password`, with no error message. Fix: add the `!` character to the list of known limitations.

**3. The pidfile location.** `%SCRIPT_DIR%coturn.pid` writes to `platform_scripts/cmd/`. The pattern `**/platform_scripts/cmd/*/` in `.gitignore` selects directories only, so git shows the file. Fix: use `%SCRIPT_DIR%coturn\coturn.pid`.

**4. `pushd %SCRIPT_DIR%coturn\` (common.bat:321).** A path that contains spaces causes an error. Fix: add quotation marks.

**5. `set SERVER_ARGS=%SERVER_ARGS% %1` in `:PostArgs`.** An `&` character in a server argument causes an error. Fix: use `set "SERVER_ARGS=%SERVER_ARGS% %1"`.

**6. The description says "no behaviour changes".** With `--start-turn --turn <external-server>`, a local coturn process now starts where previously none did, and `peer_options` now carries the default credentials instead of empty strings. This is the intended correction. Fix: add a line to the description.

Problem 1 needs a change or a test. Problems 2 to 6 do not prevent a merge.

ChrisMaroglou:
Thanks for the review @mcottontensor,

I addressed 5 of the 6 issues in the follow-up commits. Item 5 I'd like to push back on.

 **1. `call "!TURN_PROCESS!" !TURN_ARGS!`** 
Removed "Call".

**2. `!` in a credential** 
Added to the known limitations and to `platform_scripts/cmd/README.txt`. Confirmed that `pass!word` becomes `password`. Fixing it properly needs delayed expansion off around `:ParseArgs`, which means `endlocal`, which drops everything the parse just set.

**3. Pidfile**
Now `"%SCRIPT_DIR%coturn\coturn.pid"`.

**4 .Quoted  `pushd`** 

**5. `set "SERVER_ARGS=%SERVER_ARGS% %1"`**
 Not applied. It breaks more than it fixes:

 | Input after `--` | PR | Quoted |
 |---|---|---|
 | `--http_root="C:\a&b\www"` | preserved | aborts, exit 1 |
 | `"a&b"` | preserved | aborts, exit 1 |
 | `a^&b` | fails before `:PostArgs` | same |
 | `--player_port 8080`, `--http_root="C:\a b\www"` | ok | ok |

The problem is quote parity. cmd decides what counts as an operator by scanning the line and flipping an in-quotes flag on every `"`, and it does that after `%1` has been substituted but before anything runs. A value like `--http_root="C:\a&b\www"` arrives with the user's own quotes still attached, so on the current line the `&` sits between them and is read as text. Wrapping the assignment adds one more quote at the front, which shifts every quote after it by one position: the user's opening quote now closes the region instead of opening it, the `&` ends up outside, and cmd splits the line there. The value is lost and the script aborts.

The inputs the quoting would protect are ones where an operator arrives without quotes around it, and those never reach `:PostArgs` — they break the `%*` handoff in the entry script long before argument parsing starts.

 **6. Updated PR Description with behaviour change**

mcottontensor:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|UE5.6|[](https://github.com/EpicGames/PixelStreamingInfrastructure/pull/963)|
|✅|UE5.7|[](https://github.com/EpicGames/PixelStreamingInfrastructure/pull/964)|
|✅|UE5.8|[](https://github.com/EpicGames/PixelStreamingInfrastructure/pull/965)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport) and see the [Github Action logs](https://github.com/EpicGames/PixelStreamingInfrastructure/actions/runs/31552646639) for details

PR #921, opened 2026-06-24
Add hooks for consumer-supplied connection authentication and authorization

mcottontensor:
## Relevant components:
- [x] Signalling server
- [ ] Common library
- [ ] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [x] SFU

## Problem statement:
The signalling server intentionally ships no authentication — consumers are expected to bring their own, appropriate to their environment. Today that is harder than it should be: there is no documented place to authenticate a connection, no way to recover an authenticated identity on a connection after the upgrade, and no way to control the id a streamer registers as. The last point also means a connection can claim a streamer id before the legitimate streamer connects (id squatting), after which the legitimate streamer is silently renamed.

This PR adds policy-free extension points so a deployment can plug in its own auth without forking. It does **not** add any authentication scheme, credentials, or login flow.

## Solution
Three seams, all opt-in and backwards compatible:

- **Authenticate at the WebSocket upgrade (existing, now documented).** `streamerWsOptions` / `playerWsOptions` / `sfuWsOptions` are already passed through to the underlying `ws` server, so a consumer can supply a `verifyClient` callback to accept/reject a connection before the server sends its config message (which carries peer/TURN options). This was undocumented; it is now described as the recommended authentication point.
- **Expose the HTTP upgrade `request` on connections** (`IStreamer.request` / `IPlayer.request`). Anything a `verifyClient` (or other front door) attaches to the request — e.g. a resolved identity — is recoverable later for authorization decisions.
- **`IServerConfig.authorizeStreamerId`** — an optional hook consulted in `StreamerRegistry` when a streamer identifies itself. It receives the requesting `streamer` (with its `request`), the `requestedId`, the `sanitizedId` the registry would otherwise commit, and whether the requested id `collided`. It returns the id to commit (accept the default, or override to e.g. namespace per tenant), or `null` to reject and disconnect the streamer. This is the seam for anti-squatting / id-ownership policy.

When neither new option is supplied, behaviour is unchanged (requested id accepted, numeric suffix appended on collision).

### Changes
- `Signalling/src/StreamerRegistry.ts`: add `IStreamerIdAuthRequest` and `StreamerIdAuthorizer` types; optional authorizer applied in `onEndpointId` (accept / override / reject, with a uniqueness guard on overrides).
- `Signalling/src/SignallingServer.ts`: add `authorizeStreamerId` to `IServerConfig`, pass it to the registry, and thread the upgrade `request` into the three connection constructors.
[...]

PR #956, opened 2026-08-11
feat(signalling): issue TURN credentials per connection

marekl11:
## Relevant components:
- [x] Signalling server
- [ ] Common library
- [ ] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
`peerOptions` is built once in the `SignallingServer` constructor and sent verbatim to every peer that connects, so a TURN username and credential written there is shared by every session for the life of the deployment. Anything that opens a player WebSocket is handed it, so in practice it's public — it ends up in devtools, in packaged builds, and in whatever config file was copied to set the deployment up — and changing it means a redeploy, so in practice nobody does. This is the weakness tip 3 of `Docs/Security-Guidelines.md` already warns about.

## Solution
`IServerConfig.peerOptionsProvider` is an optional hook called once per connecting peer, returning the peer options to send to that peer. Absent, nothing changes. A provider that throws falls back to the static `peerOptions`, because a peer that never receives a config message waits forever with nothing in its log to say why. Like `authorizeStreamerId`, the library takes no view on what a credential is.

On top of it, the signalling server gains `--turn_secret` (or `--turn_secret_file`) and `--turn_ttl`, implementing the `use-auth-secret` scheme coturn and the hosted TURN services accept: username `<expiry>:<peerId>`, credential `base64(HMAC-SHA1(secret, username))`. Only `turn:`/`turns:` entries are touched and the configured object is copied rather than modified. The secret stays in the application rather than `IServerConfig`, since `GET /api/config` returns that verbatim, and it's kept out of the log dump, `--save` and the process command line.

One decision worth flagging: **only players get a time limited credential.** A streamer receives its configuration once and nothing replaces it, so an expiry there would eventually stop the stream rather than an attacker — those peers get one that outlives any plausible uptime. I'd rather that were an informed choice than mine alone; the alternative is re-sending `config` to a streamer as its credential ages, which is a new message flow and felt like more than this PR should decide.

## Documentation
`Docs/Security-Guidelines.md` gains a section on the hook and the flags, and tip 3 now points at it. It covers the two things that surprised me: that a TTL may gate only the allocation and not a session already running, and that a short TTL limits how long a leaked credential lasts but not who is issued one — the WebSocket upgrade hook is still the answer for that.

22 unit tests across the two packages, covering the hook and the credential scheme. They include a frozen known-answer vector, so a drift to base64url, hex or another digest fails rather than being recomputed the same wrong way by the test. `SignallingWebServer` had no test runner, so this adds jest to it and to `Signalling`, matching `Frontend/library`'s configuration; both were already hoisted for that package, so the lockfile diff is six lines.

With no secret set, `peerOptionsProvider` is undefined and the path is the one that was there before — I upgraded an existing deployment without a secret first and confirmed it behaved as it had. Then, against coturn 4.6.1 in `use-auth-secret` mode: a credential minted by this code allocates successfully, the retired static credential is refused with `401`, and a real stream ran through the relay from a corporate network behind symmetric NAT — where no direct path exists — surviving repeated reconnects, each minting a fresh credential.

Worth knowing if you adopt this: coturn treats `static-auth-secret` and `user=` as mutually exclusive and logs *"Shared secret overrides username/password based auth method"*, so there's no overlap window — the TURN server config and the signalling server have to change together.


mcottontensor:
## 💔 Some backports could not be created

| Status | Branch | Result |
|:------:|:------:|:------|
|❌|UE5.6|**Backport failed because of merge conflicts**<br><br>You might need to backport the following PRs to UE5.6:<br> - [feat(signalling): add hooks for consumer-supplied connection auth (#921)](https://github.com/EpicGames/PixelStreamingInfrastructure/pull/921)|
|❌|UE5.7|Backport failed because of merge conflicts|
|✅|UE5.8|[](https://github.com/EpicGames/PixelStreamingInfrastructure/pull/973)|

### Manual backport
To create the backport manually run:
```
backport --pr 956
```

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport) and see the [Github Action logs](https://github.com/EpicGames/PixelStreamingInfrastructure/actions/runs/31566659339) for details

PR #1086, opened 2026-09-15
feat(signalling): add optional streamer_token auth

s-uryansh:
## Relevant components:## Relevant components:
- [x] Signalling server
- [ ] Common library
- [ ] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
Streamer port has no auth. player_token exists for players but there's nothing equivalent for streamers anyone who can reach that port can connect and register as a streamer.

## Solution
Added `--streamer_token` and `--streamer_token_file`, same idea as player_token. Reused the existing verifier instead of writing a new one (renamed it to `createTokenVerifier` since it's shared now). Hooks into `streamerWsOptions.verifyClient` so bad tokens get rejected at the upgrade step, before a connection is even made.

Opt-in if you don't set a token nothing changes. Token can go in the query string or as a Bearer header, same as player_token. Also made sure it gets redacted from --save, logs, and the interactive config printout.

## Documentation
Updated README with the new flags, and updated Security-Guidelines.md since it used to flat out say streamer connections aren't authenticated that's not true anymore if you set the token.

## Test plan and compatibility
  Added tests covering the Commander option parsing and token-file loading paths, including:
  - Authentication disabled when no token is configured
  - Missing and incorrect tokens rejected with `401`
  - Correct token accepted
  - Invalid configuration values rejected
  - Missing, empty, and valid token files
  - Token-file whitespace trimming
  - Configuration-log redaction

  Updated the existing player-token tests for the shared verifier rename without removing any assertions.

  Verified on Node.js 22.14.0:

  - SignallingWebServer test suite passes
  - Build passes
  - Lint passes
  - Live WebSocket upgrade checks return the expected authorization results

Nothing else changes if you're not using the new option.


s-uryansh:
# Why is there no review

mcottontensor:
## Review

Nice work — the auth path itself is correct. I checked out the branch, rebuilt `Common`/`Signalling`/`SignallingWebServer`, and ran a live smoke test with `--streamer_token`: streamer with no token → 401, wrong token → 401, correct token → accepted; player and SFU ports unaffected; the refusal logs the peer address only; `--log_config` prints `"streamer_token": "<redacted>"`. `npm run lint` and `tsc --noEmit` are clean and 65/65 Jest tests pass. The feature is genuinely opt-in, and the docs' "SFU connections are unaffected" claim checks out (the SFU connects on `sfu_port`).

No correctness bugs found. The findings below are consistency/robustness issues.

---

**1. `SignallingWebServer/src/streamerAuth.ts:59` — a numeric token aborts startup, unlike `player_token`.**

`configureStreamerToken` rejects any non-string token, so `{"streamer_token": 12345678}` in `config.json` refuses to boot:

```
error: A streamer token was configured but is invalid; refusing to start with an open streamer port.
Error: Invalid streamer_token.
```

whereas `{"player_token": 12345678}` deliberately coerces and warns `player_token was given as a number and is being used as the text "12345678". Quote it in the config file...`. `index.ts:367-379` has an explicit comment noting JSON has one number type and that this case is expected from hand-written configs. Since both the PR description and `Docs/Security-Guidelines.md` say streamer tokens work "the same" as player tokens, an operator who copies the player pattern gets a server that won't start — and the error message never suggests quoting it. It fails closed, so this isn't a security hole, but it's worth either coercing numbers like the player path does or saying "must be a quoted string" in the message.

**2. `SignallingWebServer/src/index.ts:406` — bare `catch {` discards the cause.**

The handler unconditionally reports "A streamer token was configured but is invalid". `configureStreamerToken` also calls `createTokenVerifier`, so any failure originating there (or from future validation added inside) gets relabelled as an invalid token and the original message/stack is lost. Every other validation in this file logs a cause-specific message. Either narrow the `try` to the type check, or include the caught error in the log line.

**3. `SignallingWebServer/src/streamerAuth.ts:45` — throw with no preceding `Logger.error`.**

`loadStreamerTokenFile` throws `Error('Invalid streamer_token_file.')` without logging first. Every other config validation in `index.ts` — including `readSecretFile` two functions above — logs first, precisely because the throw beats winston's flush. With `{"streamer_token_file": true}` in `config.json` the operator gets only a bare stack trace, with no line naming the offending option and nothing in the log file.

**4. `SignallingWebServer/src/streamerAuth.ts:27` — `readSecretFile` landed in a streamer-specific module.**

It's the shared reader for `turn_secret_file` and `player_token_file` too, so `index.ts:311` and `index.ts:316` now import a generic helper from `streamerAuth.ts`. This same PR moved `SECRET_OPTIONS`/`redactConfig` into `Utils.ts`, which looks like the consistent home for this as well. The move also dropped the doc comment explaining *why* it trims (`echo $SECRET > file` leaves a newline) and *why* an empty file throws (an empty secret silently reads as "not configured") — that's the kind of rationale this file tends to keep.

**5. `SignallingWebServer/src/streamerAuth.test.ts:7` — wilbur's unit tests now require `Signalling` to be built.**

The suite imports `Logger` as a runtime value (for `jest.spyOn`) from `@epicgames-ps/lib-pixelstreamingsignalling-ue5.8`. On a stale `Signalling/dist` this fails with `Cannot find module ... from '../Signalling/dist/cjs/Logger.js'` while `playerAuth.test.ts` and `turnCredentials.test.ts` still pass. Note that `playerAuth.ts:6-12` declares its own `VerifyClient` type specifically to avoid depending on the lib. CI builds `Signalling` first (`.github/workflows/healthcheck-libraries.yml:46-52`) so CI is unaffected, but `cd SignallingWebServer && npm test` on a fresh or stale tree now fails with an unrelated-looking error. Dropping the `Logger.error` spy removes the dependency — those error paths are already asserted via the thrown message.

**6. `Docs/Security-Guidelines.md:52` — no guidance on how a streamer presents the token.**

The "token has to reach the signalling WebSocket" guidance that follows only covers browser players (`ss=`, `SignallingServerUrl`). There's no equivalent for the UE streamer (`-PixelStreamingURL="ws://host:8888?token=..."`), which is the only client the new flag applies to — so an operator who sets `--streamer_token` and then can't get the streamer to connect has nothing to go on. Relatedly, the `--streamer_token` help text (`streamerAuth.ts:13`) drops the "refused at the HTTP upgrade with 401" clause that `--player_token` carries, which makes the two README rows read asymmetrically.

---

Nit, non-blocking: `streamerAuth.test.ts` doesn't pass `prettier --check` (long `expect(...)` lines). `turnCredentials.test.ts` already doesn't either and there's no format gate in CI, so this is cosmetic.

